Issue #11, opened 2020-03-06
Storybook: Stage Details


dpfernandes:
When I click on the Stage Details accordion option on Add Crop, I want to be able enter the optimal and critical values for each of the stages/sensors

Create buttons for each STAGE that allow to change to the different STAGE FORMS and allow to enter the critical and optimal values for each of the sensors - Temperature, Humidity, PH, EC and Water Level.

<img width="288" alt="Screenshot 2020-03-06 at 16 37 12" src="https://user-images.githubusercontent.com/6642037/76097911-cb78a200-5fc8-11ea-96c5-ef1dcf45d042.png">

Note: Create the components with the appropriate style. Please check the wireframes https://www.figma.com/file/BGeghDJzrIq4rq8QsbGa5p/Seasony-final-flow?node-id=0%3A1

PR #19, opened 2020-03-09
Storybook: Stage Details [Issue #11]

kfazinic:
- reusable `Button` component (see different variants in Storybook)
- `SectionHeader` component for the dashboard accordion
- `StageDetails` component as in Issue #11


src/client/components/button/button.component.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import './button.style.css';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import theme from '../../theme';
+
+export default function Button({ children, type, size, onClick, toggled }) {
```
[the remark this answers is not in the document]
kfazinic: I'll be removing mine as soon as she merges her code. 👍

kfazinic:
@zkwsk would you mind taking another look at this? :) I got the json file with crop default values from Daniel and changed the component so it can display it (and later edit it).

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+  useEffect(() => {
+    setDataForStage(
+      cropData
+        ? cropData.filter(
+            (dataPoint) =>
+              parseInt(dataPoint.fk_crop_stage_id, 10) === selectedStageNumber,
+          )
+        : [],
+    );
+  }, [selectedStageNumber, cropData]);
```
bytewiz:
```javascript
useEffect(() => {
    setDataForStage(
      cropData
        ? cropData.filter(
            (dataPoint) =>
              parseInt(dataPoint.fk_crop_stage_id, 10) === selectedStageNumber,
          )
        : [],
    );
  }, [selectedStageNumber, cropData]);
```

So `cropData` should not be able to be anything else but an array the way I see it so don't see a reason to have the extract check. and if you need it, you should rather check it using `&&` and then not update it I would say.

Also this is a bit easier to read, with a variable explaining the filter 👍 

```javascript
useEffect(() => {
   const selectedStages = cropData.filter(({ fk_crop_stage_id }) => parseInt(fk_crop_stage_id, 10) === selectedStageNumber)
    setDataForStage(selectedStages);
  }, [selectedStageNumber, cropData]);
```

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+          const updatedDataPoint = {
+            ...dataPoint,
+            [`${level}_value`]: valueEntered,
+          };
+          return updatedDataPoint;
+        }
+        return dataPoint;
+      });
+    });
+  }
```
bytewiz: I would consider making this a separate function instead of inline 👍 as it is fairly functional

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+      cropData
+        ? cropData.filter(
+            (dataPoint) =>
+              parseInt(dataPoint.fk_crop_stage_id, 10) === selectedStageNumber,
+          )
+        : [],
+    );
+  }, [selectedStageNumber, cropData]);
+
+  function handleInputChange(e, level, parameter) {
```
bytewiz: `parameter` seems very generic, could basically be anything? could we make it more descriptive 🤔

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+          {levels.map((level) => {
+            return (
+              <div
+                className={`${level}-column`}
+                key={`stage-${selectedStageNumber}-${level}-column`}
+              >
+                {Object.keys(parametersAndLabels).map((parameter) => {
+                  const dataForParameter = dataForStage.filter((dataPoint) => {
+                    return dataPoint.parameter === parameter;
+                  });
```
bytewiz: If you only want the first and one item in the array you might as well use `.find()` rather than `.filter()`, then you can avoid the `[0]`

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+                  const value = dataForParameter[0]
+                    ? dataForParameter[0][`${level}_value`]
+                    : '';
+                  const name = `${level}-${parameter}-${
+                    stages[selectedStageNumber - 1].name
```
bytewiz: is `selectedStageNumber` an index? if yes, maybe it should be renamed to `selectedStageIndex` to make it easier 👍

src/client/components/stage-details/stage-details.stories.js
```diff
@@ -0,0 +1,40 @@
...
+const cropID = 1;
+
+const parametersAndLabels = {
+  temperature: 'Temperature (°C)',
+  humidity: 'Humidity (g/m3)',
+  ph: 'PH',
+  ec: 'EC (ppm)',
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  water_level: 'Water level (cm)',
+};
```
bytewiz: I am not very happy with variable or function names that include `And` 😁 

In my opinion, these are not parameters and labels but rather `units` so I would rename it like this and make it more clear it is something constant:
```javascript
const UNITS = {
  TEMPERATURE: 'Temperature (°C)',
  HUMIDITY: 'Humidity (g/m3)',
  PH: 'PH',
  EC: 'EC (ppm)',
  // eslint-disable-next-line @typescript-eslint/camelcase
  WATER_LEVEL: 'Water level (cm)',
};
```


The last thing is I think this should be defined in a place where it can be reusable and not in the story 👍

src/client/components/section-header/section-header.component.js
```diff
@@ -0,0 +1,23 @@
...
+import './section-header.style.css';
+
+export default function SectionHeader({ children, onClick, tabIndex }) {
+  return (
+    <div
+      className="section-header"
+      role="button"
+      onClick={onClick}
+      onKeyDown={onClick}
+      tabIndex={tabIndex}
```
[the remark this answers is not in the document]
moonflare: > Does this do anything at all? don't think `tabIndex` is a valid prop on a `div` 🤔

It is valid. It relates to[ tab navigation](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

src/client/components/section-header/section-header.component.js
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+import './section-header.style.css';
+
+export default function SectionHeader({ children, onClick, tabIndex }) {
+  return (
+    <div
+      className="section-header"
+      role="button"
+      onClick={onClick}
+      onKeyDown={onClick}
```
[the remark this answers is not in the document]
kfazinic: It's an accessibility thing, eslint doesn't let me not have both: 
> Visible, non-interactive elements with click handlers must have at least one keyboard listener.eslint(jsx-a11y/click-events-have-key-events)

➡️ For reference, 1st paragraph here as well as the link in it: https://stackoverflow.com/a/54612218/7547895

I guess I could have created it using a `button` instead of a `div`. It's not too late for that if there would be less confusion...

src/client/components/section-header/section-header.component.js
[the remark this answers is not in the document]
kfazinic: What @moonflare said. It's eslint again...
> Elements with the 'button' interactive role must be focusable.eslint(jsx-a11y/interactive-supports-focus)

src/client/components/section-header/section-header.stories.js
```diff
@@ -0,0 +1,37 @@
...
+          backgroundColor: '#ebecf0',
+          display: 'flex',
+          flexDirection: 'column',
+          height: '20em',
+          justifyContent: 'space-around',
+        }}
+      >
+        {sectionTitles.map((title, index) => {
+          return (
+            <SectionHeader onClick={action('header clicked')} tabIndex={index}>
```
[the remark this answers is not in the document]
kfazinic: Hmm, I pass this `onClick` both to the `div`'s `onKeyDown` and `onClick`  through props. It kind of needs to be in there (see previous comments). :D 

But, as I said, if this will work better as a button, I have no problem with rewriting it. 👍

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+import './stage-details.style.css';
+import SectionHeader from '../section-header/section-header.component';
+import Button from '../button/button.component';
+
+function capitalizeString(str) {
+  return str[0].toUpperCase().concat(str.slice(1, str.length));
+}
+
+export default function StageDetails({
+  cropID,
```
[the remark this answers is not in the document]
kfazinic: Ty! ✔️

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+  useEffect(() => {
+    setCropData(
+      data.filter((dataPoint) => parseInt(dataPoint.fk_crop_id, 10) === cropID),
+    );
+  }, [cropID, data]);
+
+  // Select the last stage to display as default
+  const [selectedStageNumber, setSelectedStageNumber] = useState(
+    Math.max(...stages.map((s) => s.id)),
+  );
```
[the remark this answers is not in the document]
kfazinic: Good point! ✔️

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
+import React, { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import './stage-details.style.css';
+import SectionHeader from '../section-header/section-header.component';
+import Button from '../button/button.component';
+
+function capitalizeString(str) {
+  return str[0].toUpperCase().concat(str.slice(1, str.length));
+}
```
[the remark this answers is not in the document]
kfazinic: Wasn't aware I could do that, nice, thanks! 👏 ✔️

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+  return (
+    <div className="stage-details-container">
+      {/* Section header */}
+      <SectionHeader onClick={() => null} tabIndex={0}>
+        Stage Details
+      </SectionHeader>
+
+      <form
+        className="stage-values-form"
+        onSubmit={(e) => saveUpdatedCropData(e)}
```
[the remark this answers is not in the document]
kfazinic: Ty! ✔️

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+
+  function saveUpdatedCropData(e) {
+    e.preventDefault();
+    /* TO DO: SAVE UPDATED cropData */
+  }
+
+  return (
+    <div className="stage-details-container">
+      {/* Section header */}
+      <SectionHeader onClick={() => null} tabIndex={0}>
```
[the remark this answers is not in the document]
kfazinic: Technically it should hide or display the section below the header (it's going to resemble an accordion with one of the sections always open, I assume; see [here](https://www.figma.com/file/BGeghDJzrIq4rq8QsbGa5p/Seasony-final-flow?node-id=0%3A1)). I didn't write that function yet...

`tabIndex` could then be `0, 1, 2, 3` for the four headers so the user can move between them using the keyboard. I provided it here because it's required.

src/client/components/stage-details/stage-details.component.js
kfazinic: Ty! ✔️

src/client/components/stage-details/stage-details.component.js
kfazinic: True, ty! ✔️

src/client/components/stage-details/stage-details.component.js
kfazinic: "parameter" is the name Seasony use in their DB ([diagram](https://dbdiagram.io/d/5e79e58e4495b02c3b88b043)) for their measures (temperature, water_level, EC, PH, humidity), so I thought this would be the clearest way of naming it. 🤔

src/client/components/stage-details/stage-details.component.js
kfazinic: Nice, thank you! ✅

src/client/components/stage-details/stage-details.stories.js
kfazinic: Those are all great points, thanks! 💯 

If it's not much of an issue, a name like `PARAMETERS` could be better for clarity since that's what the company calls them (see [comment](https://github.com/HackYourFuture-CPH/FP-class11/pull/19#discussion_r398820603)). 

I understand it's good to write the constants using uppercase, but in this case, the keys correspond to the actual values in Seasony's data output, so perhaps converting them to lowercase any time they're used would add to the confusion? E.g. from [here](https://github.com/HackYourFuture-CPH/FP-class11/pull/19/files#diff-b9e92b9fb13172277b5dad8d319f74eb):

```
  {
    "id": "1",
    ....
    "parameter": "temperature",
    ...
  },
  {
    "id": "2",
    ....
    "parameter": "humidity",
    ...
  },
```

I'll relocate the object to inside the component's file. 👍

src/client/components/stage-details/stage-details.component.js
kfazinic: Done ✅

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return dataPoint;
+      });
+    });
+  }
+
+  function saveUpdatedCropData(e) {
+    e.preventDefault();
+    /* TO DO: SAVE UPDATED cropData */
+  }
```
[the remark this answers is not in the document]
kfazinic: Yes, I added an `onChange` prop to the component which it will be receiving from the container. 

Now it's 
```
  function saveUpdatedCropData(e) {
    e.preventDefault();
    onChange(cropData);
  }
```

src/client/components/stage-details/stage-details.component.js
```diff
@@ -0,0 +1,173 @@
...
+
+  return (
+    <div className="stage-details-container">
+      {/* Section header */}
+      <SectionHeader onClick={() => null} tabIndex={0}>
+        Stage Details
+      </SectionHeader>
+
+      <form className="stage-values-form" onSubmit={saveUpdatedCropData}>
```
senner007: There are at least 5 - 10 components nested/hidden in this GIGANTIC!!! function. 
Please, remember the "React way": https://reactjs.org/docs/thinking-in-react.html

This might not be obvious now, but later when you are working on a project where you have 200 components written like this, and the bug´s start to appear (and they will!), the whole thing will collapse. 
Remember what I said 😇, web applications a simple computer programs, but usually with enormous amount of code. This means that your most important considerations, are **readability and maintainability**. If you write code only to be seen by yourself... then do whatever, but when collaborating, everything you write has be understood be everyone else also, and everyone else is busy doing their own stuff. 

1. Whenever you are returning nested divs, its safe to assume that these should be replaced by components.
2. Indentation and inline logic like map() and Object.keys() make the code harder to reason about. This logic should be hidden away behind a layer of abstractions.
3. Designers and other programmers coming from other languages will have a hard time interpreting code written in an imperative style.  A declarative style, on the other hand, which btw, is also the "thinking in react way", serves to explain to the recipient _what_ the  code does and not _how_ javascript works.

src/client/components/section-header/section-header.component.js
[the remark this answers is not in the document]
bytewiz: Ah okay cool 👍

src/client/components/section-header/section-header.stories.js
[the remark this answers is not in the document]
bytewiz: Alright if that is the lint rules we go with then let's allow it 👍

src/client/components/stage-details/stage-details.component.js
bytewiz: Alright, if that makes sense to you then that is good 👍 I can live with that, I am just not a fan of generic namings but if the whole picture makes sense let's keep it as is 👍

senner007:
Yes agree, please don't use "handleInputChange" name.
What input?

src/client/components/stage-details/stage-details-subcomponents/value-input.component.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function ValueInput({
```
senner007: These components still don't reveal any information about what they do and what they contain.
What is parameter and parameters?
What is level, handleValueInput?
What title, what value, what name?
What is ValueInput?

senner007: Would it be correct to say that PARAMETERS is something like cropEnvironmentData?

senner007: Or cropEnvironmentSensors maybe

kfazinic:
@senner007  I tried to implement it similarly to your 2nd suggestion. I hope it's okay, can't see how to break it up further. Naming could be improved, but I'm out of ideas right now. BTW, the data is editable, not just displayed.

senner007:
If this works for you, then I'm good with it also. 👍

PR #23, opened 2020-03-11
Storybook: Login Form component [Issue #1]

### LINKED ISSUE #1: Storybook: Login Form component
When I click on the Login button, I want to be able to login using a login form so that I can sign in using email and password. A login error should be indicated on the form if an incorrect email or password is entered.

The login should be a form component and should have the following functionality:

- Seasony logo & Title
- Input text box for Email & Password
- Login error label & input error indication
- Button for Login & Forget password

Screenshot 2020-03-06 at 11 40 46Screenshot 2020-03-06 at 11 41 35

Note: Create the components with the appropriate style. Please check the wireframes https://www.figma.com/file/BGeghDJzrIq4rq8QsbGa5p/Seasony-final-flow?node-id=0%3A1

cecastosic:
resolves #1


src/client/components/container/container.component.js
```diff
@@ -0,0 +1,69 @@
...
+  );
+}
+
+Container.defaultProps = {
+  error: false,
+};
+
+Container.propTypes = {
+  title: PropTypes.string.isRequired,
+  error: PropTypes.bool,
```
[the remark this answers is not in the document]
zkwsk: But where will you get the contents of the error notification from? The state can be controlled based on whether the error variable is empty or not.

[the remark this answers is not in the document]
cecastosic: Shouldn't that be defined in notification component then, not here?

src/client/components/button/button.component.js
```diff
@@ -0,0 +1,38 @@
...
+export default function Button({ children, type, size, onClick, toggled }) {
+  const btnClass = classNames({
+    btn: true,
+    'btn-primary': type === 'primary',
+    'btn-secondary': type === 'secondary',
+    'btn-danger': type === 'danger',
+    'btn-toggle': type === 'toggle',
+    'btn-large': size,
+    toggled, // true if the toggle button is pressed and held down
+  });
```
zkwsk: I prefer to add "static" classes that are not generated based on logic to be declared outside the object like this:

```suggestion
  const btnClass = classNames('btn', {
    'btn-primary': type === 'primary',
    'btn-secondary': type === 'secondary',
    'btn-danger': type === 'danger',
    'btn-toggle': type === 'toggle',
    'btn-large': size,
    toggled, // true if the toggle button is pressed and held down
  });
```

src/client/components/button/button.style.css
```diff
@@ -0,0 +1,54 @@
...
+  font-size: 1.7em;
+  line-height: 1.2em;
+  padding: 0.3em 3em;
+}
+
+.btn-wrapper .btn-toggle {
+  background: var(--light-grey);
+  border-radius: var(--border-radius-low);
+  box-shadow: var(--embossed-shadow-medium);
+  color: var(--font-color);
```
zkwsk: Very nice 👍

src/client/components/footer/footer.style.css
```diff
@@ -1,5 +1,13 @@
...
+  margin: 0;
+  padding: 0;
+  color: #3b4349;
+  font-family: Proxima Nova, sans-serif;
+}
+
 .footer {
-  /* TODO: background-color to be deleted later when we use global style */
+  /* TODO: background-color to be deleted later when we use use global style */
   background-color: #ebecf0;
```
zkwsk: Haha... I think it was a typo but you added another "use". Can't you put the background color of the footer in the theme file now? And remove the TODO, obviously.

src/client/components/input-login/input-login.style.css
```diff
@@ -0,0 +1,43 @@
...
+
+.login input.error + i {
+  background: var(--danger-color);
+  color: var(--white);
+  padding-bottom: 1em;
+}
+
+.login input.error + i path {
+  color: var(--white) !important;
+}
```
zkwsk: Looks quite hacky with the `!important`. Why is this necessary? Are you styling an svg?

src/client/components/link/link.style.css
```diff
@@ -0,0 +1,6 @@
+.link {
+  color: #666;
```
zkwsk: Shouldn't the color go in the theme?

src/client/components/login/login.component.js
```diff
@@ -0,0 +1,76 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  useEffect(() => {
+    if (errorNotifications) {
+      if (errorNotifications === 'email')
+        setErrorMessage('This is error message for wrong email address');
+      else if (errorNotifications === 'password')
+        setErrorMessage('This is error message for wrong password.');
+      else setErrorMessage('Something went wrong. Try again later.');
+    }
+  }, [errorNotifications]);
```
zkwsk: Looks like you're trying to make this a container component without it actually being the case.

You could make this a container, but if you do so you can't put it in storybook.

I think you're planning on calling the api in the useEffect hook, but I don't think it belongs here. I think there is yet another component above this one which will have the state. That component will be responsible for calling the API and passing data (or errors) from the API down to this component via props.

Remember the hierarchy I did:

    - LoginContainer (container)
      - Login (presentational, this component)
        - InputLogin
        - Button
        - etc

Now it looks like you are trying to do:

      - Login (container, this component)
        - InputLogin
        - Button
        - etc

But your container has props and `useEffect()` at the same time. If a container component calls an API it will usually get all of its state from there, not from props. There can still maybe be a few props related to theming or similar, but you would not have a component where you can give it data via props and it will also get data from an API call. If you do that the component will be functioning in very different ways when you use it in Storybook (with props) and in the real application (with an API call). The point of Storybook is to make as realistic an environmnet for the component as possible, but still isolated from the outside (isolated from APIs that may not be available and other side effects).

src/client/components/login/login.stories.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import Login from './login.component';
+import { action } from '@storybook/addon-actions';
+import { withKnobs, text, select } from '@storybook/addon-knobs';
+
+export default {
+  title: 'ContainerLayouts/Login',
```
zkwsk: What do you plan to put in the category "ContainerLayouts"? I think it's a bit of a strange name. You can just scratch the category and give it a title for now. We can group things later.

```suggestion
  title: 'Login',
```

src/client/components/notification/notification.stories.js
```diff
@@ -0,0 +1,17 @@
...
+  component: Notification,
+  decorators: [withKnobs],
+};
+
+export const Error = () => {
+  const txt = text(
+    'Error text',
+    'This is error message for wrong email address.',
+  );
+  return <Notification text={txt} />;
```
zkwsk:
```suggestion
  const errorText = text(
    'Error text',
    'This is error message for wrong email address.',
  );
  return <Notification text={errorText} />;
```

I'm nitpicking a bit, but really don't like this types of "deliberate misspellings" because a given name is already used for something else. Just make longer names and let your editor autocomplete for you, it's easier to read :)

zkwsk: If you're lazy this is also an option, but I like the pattern of giving knobs their own variables.

```suggestion
  return <Notification text={
     text(
      'Error text',
      'This is error message for wrong email address.',
    )
  } />;
```

src/client/components/login/login.stories.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const notifications = {
+  None: null,
+  Email: 'email',
+  Password: 'password',
+  General: 'general',
+};
+
+export const LoginContainer = () => {
```
zkwsk: Since it's not a container component let's not call the story that.

I realize that you import the Login component under the name Login, so you can't use that, which I find really annoying with the new way of doing storybook, but how about:

```suggestion
export const LoginForm = () => {
```

src/client/components/footer/footer.style.css
cecastosic: Not mine code, but yes...

src/client/components/notification/notification.stories.js
cecastosic: Isn't "txt" acronym for text, as "btn" is for button, but yes I can put longer name.

.env.example
```diff
@@ -1,24 +0,0 @@
...
-SECRET_KEY="[FILL RANDOM LONG AND VERY SECRET SECRET KEY FOR USE WITH AUTHENTICATION SERVICE HERE]"
-
-# Your Client will run on this port
-CLIENT_PORT="3000"
-
-# Enable or disable opening a new browser window every time you run webpack development server
-OPEN_BROWSER="true"
-
-# Turn this on for Knex to print SQL queries (only for debugging, don't use on production)
-# DEBUG=knex:query
```
zkwsk: Woops. Please don't delete the .env.example file. This one serves as a template so it's easy to create a .env file which is the file where you will have all your local secrets and which is not checked into git. This one should be checked in so we can share the template.

PR #46, opened 2020-03-17
created tables for users, roles, user_roles using knex

### LINKED ISSUE #58: Knex Migrations for Users, Roles and User_Roles
Create the Knex Migrations in order to create the DB schema for the following tables:

Table roles 
Table users 
Table user_roles

MirSzabo:
tables 1, 2, 3 created
resolves #58

### DIFF
--- a/src/server/migrations/20200317214852_add_users_schema.js
+++ b/src/server/migrations/20200317214852_add_users_schema.js
@@ -1,0 +1,26 @@
+exports.up = function(knex) {
+  return knex.schema.createTable('users', function(table) {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
+    table.string('name').notNullable();
+    table
+      .string('email', 60)
+      .unique()
+      .notNullable();
+    table
+      .timestamp('updated_at')
+      .defaultTo(knex.fn.now())
+      .notNullable();
+    table
+      .timestamp('created_at')
+      .defaultTo(knex.fn.now())
+      .notNullable();
+    table.timestamp('deleted_at');
+  });
+};
+
+exports.down = function(knex) {
+  return knex.schema.dropTable('users');
+};
--- a/src/server/migrations/20200317230828_add_user_roles_schema.js
+++ b/src/server/migrations/20200317230828_add_user_roles_schema.js
@@ -1,0 +1,37 @@
+exports.up = function(knex) {
+  return knex.schema.createTable('user_roles', function(table) {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
+    table
+      .integer('fk_role_id')
+      .unsigned()
+      .notNullable();
+    table
+      .integer('fk_user_id')
+      .unsigned()
+      .notNullable();
+    table
+      .timestamp('updated_at')
+      .defaultTo(knex.fn.now())
+      .notNullable();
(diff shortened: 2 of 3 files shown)


src/server/migrations/20200317230828_add_user_roles_schema.js
```diff
@@ -0,0 +1,37 @@
...
+      .timestamp('updated_at')
+      .defaultTo(knex.fn.now())
+      .notNullable();
+    table
+      .timestamp('created_at')
+      .defaultTo(knex.fn.now())
+      .notNullable();
+    table.timestamp('deleted_at');
+    table
+      .foreign('fk_role_id', 'fk_role_id')
```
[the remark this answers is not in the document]
MirSzabo: if we do not specify the foreign key name then the name is created like this:
user_roles_fk_role_id_foreign
I wanted to have name fk_role_id_foreign, that's why it is twice
but we can change it later based on requirements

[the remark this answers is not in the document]
dpfernandes: That looks fine, the example on knexjs documentation has a slightly different syntax that is easier to read because it doesn't repeat the name of the field (.foreign('fk_role_id', 'fk_role_id') but if you tested your syntax and it works go ahead. If you have any problem with it use this example http://knexjs.org/#Schema-foreign

src/server/migrations/20200317214852_add_users_schema.js
```diff
@@ -0,0 +1,29 @@
+exports.up = function(knex) {
+  return knex.schema.createTable('users', function(table) {
+    table
+      .increments('id')
+      .notNullable()
+      .primary();
```
mahieakhtar: Please add another field here ("uid" string [null])
Please follow the updated DB tables changes.

ricardoaguiar: The uid should not be null. Firebase assigns a uid for each user.

mahieakhtar: Yes, uid should not be null but according to Daniel will keep it null now, may be later it will be changed.

ricardoaguiar: We will be hard coding the user data as per Daniel.

mahieakhtar: @Daniel please clear this as yesterday we updated new DB as per your suggestion.

senner007:
Is it necessary with a users table? The users are already stored in a firebase table

ricardoaguiar:
@senner007 We need the firebase uid data to be on the users table, to validate the authentication and user roles. We cannot assign a user role on firebase.

senner007:
Yes you can:
```
// Set admin privilege on the user corresponding to uid.

admin.auth().setCustomUserClaims(uid, {admin: true}).then(() => {
  // The new custom claims will propagate to the user's ID token the
  // next time a new one is issued.
});
```
If you create a users table in addition to the one you have in firebase, you will have users state duplicated, which might lead to bugs

senner007:
Setting roles is also besides the point. You can still have UserRoles table with a users table in mysql

dpfernandes:
Approved

PR #54, opened 2020-03-22
chore: Add DB Migrations check to GitHub Actions

moonflare:
### What changed?
- Added migrations check to the script
- Changed the name to `ESLint & Migrations Check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: ESLint Check
+name: ESLint & Migrations Check
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,55 @@
...
+      env:
+        DB_HOST: "127.0.0.1"
+        DB_PORT: 33306
+        DB_NAME: "test"
+        DB_USER: "root"
+        DB_PASSWORD: "password"
+    - name: Run DB Migrations Rollback
+      run: npm run db:clean
+      env:
+        DB_HOST: "127.0.0.1"
```
senner007: maybe put these vars in secrets if they are the same

zkwsk: We can do that, but I see nothing critical about having it hard coded and we need the tests running migrations, so I'll merge this one and we can add gh secrets later.

PR #55, opened 2020-03-22
fix: add rule to only warn on camelcase

zkwsk:


zkwsk:
@zenkaeser let's try to schedule a call and figure out what happened here? Looks like you got more changes than intended in there and I can't really figure out what is supposed to be there.

Issue #5, opened 2020-03-06
Storybook: Crop Status - Info Box


dpfernandes:
When I click on the Dashboard button on the sidebar menu, I want to be able to see summary of the Crop Status on a small info box.

The info box should show the following data:

**Status:**                                      `ON TIME`  or `CHECK PH` or `FINISHED` or `CANCELED` - change color
**Days left to harvest:**                `in X days `
**Days left to end the project:**  `X days `
**Current stage:**                         `SEEDING - X DAYS` or  `PROPAGATION - X DAYS` or `MATURITY - X DAYS`
**Production Start Date:**            `DD/MM/YYYY`
**Production End State:**             `DD/MM/YYYY`

<img width="309" alt="Screenshot 2020-03-06 at 16 00 36" src="https://user-images.githubusercontent.com/6642037/76094913-b5b4ae00-5fc3-11ea-8546-ca58bc8bb58d.png"> <img width="320" alt="Screenshot 2020-03-06 at 16 00 48" src="https://user-images.githubusercontent.com/6642037/76094917-b6e5db00-5fc3-11ea-89a0-357dd96bf3af.png">

Note: Create the components with the appropriate style. Please check the wireframes https://www.figma.com/file/BGeghDJzrIq4rq8QsbGa5p/Seasony-final-flow?node-id=0%3A1

dpfernandes:
<img width="402" alt="Screenshot 2020-04-12 at 13 36 44" src="https://user-images.githubusercontent.com/6642037/79068011-eaeba480-7cc3-11ea-89f6-413cc5964e29.png">
<img width="391" alt="Screenshot 2020-04-12 at 13 36 36" src="https://user-images.githubusercontent.com/6642037/79068013-ec1cd180-7cc3-11ea-8542-136819ce315b.png">

PR #67, opened 2020-03-25
Created the Storybook: Crop Status - Info Box

zenkaeser:
Issue link: https://github.com/HackYourFuture-CPH/FP-class11/issues/5

Created a new branch with the issue title.


src/client/components/card/crop-summary/crop-summary.style.css
```diff
@@ -0,0 +1,50 @@
...
+  margin-top: -30px;
+  display: inline-block;
+}
+
+p {
+  margin: 0;
+}
+
+.check-ph {
+  background: #f27eb1;
```
gizemcandemir: Now that it’s available, we should use all colors from theme.css.

src/client/components/card/status-card02/status-card02.style.css
```diff
@@ -0,0 +1,23 @@
+.check-ph {
+  background: var(--pink);
```
gizemcandemir, line 2: we should use the logical names from theme.css instead of color names. like this:
`background: var(--danger-color);`

src/client/components/card/status-card02/status-card02.style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.finished {
+  background: var(--light-blue);
```
gizemcandemir, line 6: `background: var(--interaction-color);`

src/client/components/card/status-card02/status-card02.style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.on-time {
+  background: var(--green);
```
gizemcandemir, line 10: `background: var(--primary-color);`

src/client/components/card/status-card02/status-card02.style.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.cancelled {
+  background: var(--pink);
```
gizemcandemir, line 14: `background: var(--danger-color);`

src/client/index.css
```diff
@@ -3,7 +3,7 @@
 
 * {
   box-sizing: border-box;
-  color: #3b4349;
+  color: var(--dark-blue);
```
gizemcandemir, line 6: `background: var(--font-color);`

src/client/index.css
```diff
@@ -3,7 +3,7 @@
...
   font-family: Proxima Nova, 'Lato', 'Montserrat', sans-serif;
   font-size: 16px;
   margin: 0;
```
gizemcandemir, line 9: I think these will come from index.css for all the components, so we can remove them here.

src/client/index.css
gizemcandemir: Oh sorry this file is actually index.css 🙈 I think you shouldn't edit this file in your PR. 
Maybe that's what Niels means on his comment. Your PR should only have the changes about your task. Otherwise it will be really complicated to solve merge conflicts with other PRs and develop branch.

gizemcandemir:
By the way, the folder structure doesn't seem clear to me.. We probably should not have a component named `status-card02`. Can you rename it to `status-card`? If you think it's confusing since we have another component out under `components`, I think you should actually overwrite that one. Because that was created for demonstration purpose in the beginning and we don't need to keep it as it is. Also we don't want nested component folders. As far as I know, all components should be under main `component` folder.

Issue #76, opened 2020-03-27
Add private routing 


dpfernandes:
Build a component <PrivateRoute> that determines whether a user is authenticated and authorized to view a given page.

If the user is authorized to view a given page return a regular <Route> component showing the given page. If not, redirect the user to the login component.

You can check the previous final project for inspiration: https://github.com/HackYourFuture-CPH/ov-class09-fp/blob/develop/src/client/components/PrivateRoute.js

Route Structure
login (public)
All the other routes should be private

Stretch goal:
Pass the original requested route to the login page component and make sure to redirect the user to the right route after successful login.

PR #80, opened 2020-03-28
Firebase authentication and private routing [Issues #18 and #76]

### LINKED ISSUE #18: Firebase authentication and mysql
Firebase for authentication with backend verification.
Pretty much all of it explained in this article. 
https://dev.to/emeka/securing-your-express-node-js-api-with-firebase-auth-4b5f

Mysql is then used for the content db. 
Alternatively we could use what they call Realtime database, which is more than a db. Built-in Websockets and I believe hardbound to the authentication system, afaik 🤔

cecastosic:
resolves #18 #76 
not finished - PR made so the whole team can work together on issues 
frontend part done


senner007:
input-login and login should be in the same folder, imo

src/client/containers/login-page/login-page.js
```diff
@@ -0,0 +1,47 @@
...
+    <main>
+      <Login
+        title="Login"
+        errorNotifications={errorNotifications}
+        errorMessages={errorMessage}
+        handleChange={(event) => {
+          if (event.target.type === 'email') setEmail(event.target.value);
+          if (event.target.type === 'password') setPassword(event.target.value);
+        }}
+        loginFunc={async (event) => {
```
senner007: move logic out of view please

src/client/containers/login-page/login-page.js
```diff
@@ -0,0 +1,47 @@
...
+  const [email, setEmail] = useState('');
+  const [password, setPassword] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [errorNotifications, setErrorNotifications] = useState(null);
+
+  const history = useHistory();
+
+  return (
+    <main>
+      <Login
```
senner007: A bit too generic here. I think you should create an email and a password component.

senner007:
Plenty of single responsibility thinking here
...makes the 🐱‍👤 coder envious ...

src/client/App.js
```diff
@@ -1,14 +1,25 @@
...
+        </Route>
+        <Route exact path="/dashboard">
+          <Dashboard />
+        </Route>
         <Route exact path="/">
-          <Home />
+          <FirebaseContext.Provider value={Firebase.init()}>
+            <LoginPage />
+          </FirebaseContext.Provider>
         </Route>
```
bytewiz: I think I would wrap the entire app in the context provider as you might want to be able to display state about whether you are signed in or not in some of the pages even though they don't require authentication 👍

senner008:
I suggest that this pr. is closed. Then finish the work on the branch and open another one

src/client/App.js
cecastosic: Done

cecastosic:
@bytewiz We added private and public routes: If you are not logged in, it will show login form, if you are logged in, you can see dashboard. 
All changes related to components are there because of showing needed components on dashboard page.

zenkaeser:
Hello guys, Daniel asked me if you have already a Swagger Documentation for your endpoint on this issue if there's any?Thank you

bytewiz:
@sgalkina Only some conflicting files left it seems 👍 Please have a look at it 👍

cecastosic:
Resolved 👍

PR #81, opened 2020-03-28
Feature/chart data range buttons

### LINKED ISSUE #70: Storybook: Chart Data Range buttons stories
When I am on the Dashboard or Chart Detail, I want to be change the range of the charts, so I can see different range on the charts.

Please create the buttons stories based on the existing Button story already merged on develop

Screenshot 2020-03-26 at 08 55 38

1. Last 5 Days 
2. Last Week 
3. Last Two Weeks 
4. Last Month 
5. Custom

mamathamereddy:
resolves #70


src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,73 @@
...
+        toggled={selectedButton === buttonLabel.buttonMonth}
+      >
+        {buttonLabel.buttonMonth}
+      </Button>
+      <Button
+        type="toggle"
+        size="large"
+        onClick={(e) => {
+          e.preventDefault();
+          setSelectedButton(buttonLabel.buttonCustom);
```
ricardoaguiar, line 59:
```
   <Button
        type="toggle"
        size="large"
        onClick={(e) => {
          e.preventDefault();
```
Can you try create a single button and pass the values as props?
The code above is the same for every button.

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,33 @@
...
+      })}
+    </div>
+  );
+};
+
+ChartbarMenu.propTypes = {
+  buttonLabel: PropTypes.instanceOf(Array).isRequired,
+};
+
+export default ChartbarMenu;
```
bytewiz, line 33: It looks great! one last comment 🤓 

I would expect `buttonLabel` to be a string but I see that it is not. I would consider passing the following props
`defaultSelection` = which is optional and can be set as default state
`buttons` = which is the array of objects you map through to display all the buttons

Then change the `propTypes` accordingly 👍

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,33 @@
...
+    <div className="chartbar-button-wrapper">
+      {buttonLabel.map((obj) => {
+        return (
+          <Button
+            type="toggle"
+            size="large"
+            onClick={(e) => {
+              e.preventDefault();
+              setSelectedButton(obj.buttonId);
+            }}
```
bytewiz, line 18: Also, I don't think you need to `preventDefault` here.

If so, you could make it a one-liner 🙌

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,33 @@
...
+            {obj.buttonLabel}
+          </Button>
+        );
+      })}
+    </div>
+  );
+};
+
+ChartbarMenu.propTypes = {
+  buttonLabel: PropTypes.instanceOf(Array).isRequired,
```
bytewiz, line 30: Would be cool instead of defining `Array` to define that the array if objects only are allowed to have `buttonId` and `buttonLabel` 🥇

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,33 @@
+import React, { useState } from 'react';
+import './chart-data-button.style.css';
+import PropTypes from 'prop-types';
+import Button from '../../button/button.component';
+
+const ChartbarMenu = ({ buttonLabel }) => {
+  const [selectedButton, setSelectedButton] = useState(buttonLabel.buttonWeek);
```
bytewiz, line 7: Also according to the `story` this would be `undefined`  🙌

mamathamereddy: yes

senner008:
@bytewiz Is this ready for merge?

ricardoaguiar:
It also looks good to me

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,44 @@
...
+          </Button>
+        );
+      })}
+    </div>
+  );
+};
+
+ChartbarMenu.defaultProps = {
+  defaultSelection: '',
+  buttons: '',
```
moonflare, line 31: We need to provide the right data type here. For example, if the user doesn't provide `defaultSelection`, then on line 7, when we try to access the `id` eg `defaultSelection.id`, it will throw an error.

The same with `buttons`, if we expect it to be an array, the default value should be an empty array, otherwise `buttons.map((button) => {}` will also break if the `buttons` prop is missing.

```suggestion
  defaultSelection: {},
  buttons: [],
```

src/client/components/chart-data-buttons/chartbar-buttons/chart-data-button.component.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState } from 'react';
+import './chart-data-button.style.css';
+import PropTypes from 'prop-types';
+import Button from '../../button/button.component';
+
+const ChartbarMenu = ({ buttons, defaultSelection }) => {
+  const [selectedButton, setSelectedButton] = useState(defaultSelection.id);
```
moonflare, line 7: I would change this a bit so we don't confuse it. It actually stores the id, not the button. This is a bit more clear ⬇️ 
```suggestion
  const [selectedButtonId, setSelectedButtonId] = useState(defaultSelection.id);
```

aleksd-dfds: @mamathamereddy this is an old PR that's waiting for some changes. Could you or someone on your team implement the feedback, and merge?

ricardoaguiar:
I have pulled the branch and it looks good. Nice work.

dpfernandes:
Approved!

PR #85, opened 2020-03-30
Update add crop image component's style to use variables from theme.css

gizemcandemir:


gizemcandemir:
I will combine these changes together with footer component which was also merged before.

PR #94, opened 2020-04-01
crop_stages_endpoint

humayunadilshahzad:


src/server/api/controllers/modules.controller.crop-stages-endpoint.js
```diff
@@ -0,0 +1,24 @@
+const knex = require('../../config/db');
+
+const getCropStages = async () => {
+  try {
+    const startDate = await knex('batches')
+      .select('seeding_date')
+      .where({ id: '31' });
```
bytewiz, line 7: This should not be hardcoded, it should be passed as a parameter :)

src/server/api/controllers/modules.controller.crop-stages-endpoint.js
```diff
@@ -0,0 +1,24 @@
...
+    const stages = await knex('crop_stages').select(
+      'crop_stages.name',
+      'crop_stages.duration',
+    );
+    const date = new Date();
+    return {
+      startDate: startDate[0].seeding_date,
```
bytewiz, line 14: Instead of this, you should query `[0]` to get the first occurrence you can use `.first()` in the query for `batches`.

Also this will not behave correctly you might not get the same result everytime. Consider ordering by date or similar to make sure you get the correct value.

src/server/api/controllers/modules.controller.crop-stages-endpoint.js
```diff
@@ -0,0 +1,24 @@
...
+
+const getCropStages = async () => {
+  try {
+    const startDate = await knex('batches')
+      .select('seeding_date')
+      .where({ id: '31' });
+    const stages = await knex('crop_stages').select(
+      'crop_stages.name',
+      'crop_stages.duration',
+    );
```
bytewiz, line 11: Is this correct? Do you need to query all entries?

src/server/api/controllers/modules.controller.crop-stages-endpoint.js
humayunadilshahzad: yes it was just a test to see the behavior getting data because we are not clear . on what bases we need to select .. is it most recent batch ? what about if we have more then one batch .??

src/server/api/controllers/modules.controller.crop-stages-endpoint.js
humayunadilshahzad: what date should we consider  ?  .. i mean the last batch entered (if we have multiple batches) ? or do we have only one batch at a time ?

src/server/api/controllers/modules.controller.crop-stages-endpoint.js
humayunadilshahzad: stages I guess a gonna be same every time  ... but yes we need to get all the stages the crop is gonna pass through ... because we need to calculate total days of crop cycle ..and also the stage name to be shown under progress bar .. 
any suggestions ?

senner007:
This also contains a package-lock without new packages. Please fix that along with the merge conflicts

humayunadilshahzad:
sure Niels ... I can undo the package lock commit like I did last time as u said ... should i pull develop again to fix the merge conflicts or it should be the same way like undoing the commits

senner007:
hmm.. this might be difficult because you have several commits with package-lock changes. I think the easy way would be to completely overwrite the package-lock with the one from develop

humayunadilshahzad:
and that will fix merge conflicts as well or do we need to some other process

senner007:
That will removedthe conflict with that file only

senner007:
Next step is to pull develop and merge develop into your branch

humayunadilshahzad:
thanx Niels ...

senner008:
@humayunadilshahzad
You still need to overwrite the package-lock file with the one from develop. Simply copy and paste it in

senner008:
or 
git checkout develop package-lock.json

aleksd-dfds:
@humayunadilshahzad can you implement changes and merge?

dpfernandes:
Approved

PR #95, opened 2020-04-01
Final version of logout popup component

Nazaqat:


src/client/components/logout/logout.component.js
```diff
@@ -0,0 +1,78 @@
...
+    </div>
+  );
+};
+
+const LogoutContent = ({ children }) => {
+  return (
+    <>
+      <div className="content">
+        <p>
+          Hi {userName} <br />
```
bytewiz, line 47: `userName` should be a prop, not a hardcoded string

src/client/components/logout/logout.component.js
```diff
@@ -0,0 +1,78 @@
...
+        </p>
+      </div>
+      {children}
+    </>
+  );
+};
+
+const LogoutActions = ({ closeAction }) => (
+  <div className="actions">
+    <button className="confirm-logout" type="submit" onClick="{() => {}}">
```
bytewiz, line 58: Shouldn't you pass a "logout function" here as a prop?

src/client/components/logout/logout.component.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import Popup from 'reactjs-popup';
+import './logout.style.css';
+import PropTypes from 'prop-types';
+
+const userName = 'Servet';
```
bytewiz, line 6: Don't hardcode this :)

src/client/components/logout/logout.component.js
senner008: It's unknown by now

bytewiz: Alright, it is fine to keep like this but would rather have it as a prop and pass an "empty function" here. @senner008 what is your opinion on this? Then we shouldn't have to change the file again 🤞

senner008: you decide...

Nazaqat:
> Not sure what's going on, but you're deleting the knexfile and adding some sqlite files?

Dont know, how it happened but sqlite file has been deleted and knex.js has been added.

Nazaqat:
new branch has been created called feature/logout-component

PR #97, opened 2020-04-02
Feature/confirm_delete_batch

Ekamba:


src/client/components/Delete_Comfim/Button.component.js
```diff
@@ -0,0 +1,19 @@
...
+import Modal from './Modal';
+
+const ButtonComponent = () => {
+  const { isShowing, toggle } = Modal();
+  return (
+    <div>
+      <Button className="base-btn" type="danger" onClick={toggle}>
+        Delete BATCH
+      </Button>
+      <ModalList isShowing={isShowing} hide={toggle} onClose={toggle} />
```
kfazinic, line 14: Please see the comments I left on the `ModalList` component. 
If you applied those changes, this could just be
```
{isShowing ? <ModalList setIsShowing={setIsShowing} onDelete={() => null}/> : null}
```

src/client/components/Delete_Comfim/ModalList.component.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Button from '../button/button.component';
+
+const ModalList = ({ isShowing, hide, onClick }) => {
+  if (isShowing) {
```
kfazinic, line 6: You don't actually need to check this here, since `isShowing` is defined in the `Button` component. You can check over there whether `isShowing` is true and decide whether or not to render `ModalList` (see the last comment I left in the `Button` component file...).

src/client/components/Delete_Comfim/ModalList.component.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Button from '../button/button.component';
+
+const ModalList = ({ isShowing, hide, onClick }) => {
```
kfazinic, line 5: - You don't really need to pass `isShowing` (see the comment below)
- You don't need to pass `hide` because you can pass `setIsShowing` instead, and then replace it with `() => setIsShowing(false)`

src/client/components/Delete_Comfim/ModalList.component.js
```diff
@@ -0,0 +1,48 @@
...
+            <h4 className="text-headline">DELETE BATCH</h4>
+            <p className="p1">Are you sure?</p>
+            <p className="p2">Do you really want to delete the batch?</p>
+            <div className="buttons-wrapper">
+              <Button type="danger" size="large" onClick={onClick}>
+                Delete
+              </Button>
+              <Button type="secondary" size="large" onClick={onClick}>
+                Cancel
+              </Button>
```
kfazinic, line 38: This button just closes the modal window, so the `onClick` can just be 
```
onClick={() => setIsShowing(false)}
```

src/client/components/Delete_Comfim/ModalList.component.js
```diff
@@ -0,0 +1,48 @@
...
+            >
+              &times;
+            </span>
+          </div>
+          <div className="modal">
+            <h4 className="text-headline">DELETE BATCH</h4>
+            <p className="p1">Are you sure?</p>
+            <p className="p2">Do you really want to delete the batch?</p>
+            <div className="buttons-wrapper">
+              <Button type="danger" size="large" onClick={onClick}>
```
kfazinic, line 33: About this `onClick` function: we don't have the endpoints ready yet, but eventually when this button is clicked, it will delete the crop from the DB. `ModalList` component should receive that function through props, and you can call it `onDelete` or something similar. You can pass it as, for example, `onDelete={() => null}` since we don't yet know exactly what it will look like.

src/client/components/Delete_Comfim/Delete_Comfirm.stories.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import ButtonComponent from './Button.component';
+import { action } from '@storybook/addon-actions';
+import { withKnobs, select, boolean } from '@storybook/addon-knobs';
```
kfazinic, line 5: I think you don't need knobs here since there's no variable parts in this component...

src/client/components/Delete_Comfim/Delete_Comfirm.stories.js
```diff
@@ -0,0 +1,28 @@
...
+  <ButtonComponent
+    onClick={action('clicked!')}
+    // syntax is (name, options, default)
+    variant={select('variant', variantOptions, '')}
+    // syntax is (name, default)
+    disabled={boolean('disabled', false)}
+  >
+    Delete
+  </ButtonComponent>
+));
```
kfazinic, line 28: You're using the "old" format of writing stories. Here's what we should do instead: 

```
import React from 'react';
import DeleteCropButton from './delete-crop-button.component';

export default {
  component: DeleteCropButton ,
  title: 'Delete Crop',
};

export const DeleteCropButtonComponent = () => {
  return <DeleteCropButton />;
};
```

src/client/components/Delete_Comfim/ModalList.component.js
```diff
@@ -0,0 +1,48 @@
...
+            </div>
+          </div>
+        </div>
+      </div>,
+      document.body,
+    );
+  }
+  return null;
+};
+export default ModalList;
```
kfazinic, line 48: You're missing prop types here to let eslint know what type the props to the component should be. If you haven't used them yet, [here's how](https://www.npmjs.com/package/prop-types#usage)

src/client/components/Delete_Comfim/Button.component.js
```diff
@@ -0,0 +1,19 @@
...
+import ModalList from './ModalList.component';
+import Modal from './Modal';
+
+const ButtonComponent = () => {
+  const { isShowing, toggle } = Modal();
+  return (
+    <div>
+      <Button className="base-btn" type="danger" onClick={toggle}>
+        Delete BATCH
+      </Button>
```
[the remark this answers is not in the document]
Ekamba: Yes exactle I don't need that class anymore, I just forgot to remove when I change from my button to Ceca's button. There is a change in our project a bout crop, we no longer use that name "crop" instead we are using batch. That why I have change from crop to batch.

src/client/components/Delete_Comfim/Button.component.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import Button from '../button/button.component';
+import './Style.css';
+import ModalList from './ModalList.component';
+import Modal from './Modal';
+
+const ButtonComponent = () => {
+  const { isShowing, toggle } = Modal();
```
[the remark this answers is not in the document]
Ekamba: Thanks for your Advice

src/client/components/Delete_Comfim/ModalList.component.js
Ekamba: done

src/client/components/Delete_Comfim/ModalList.component.js
Ekamba: yes

PR #103, opened 2020-04-03
Endpoint for getting all crops

zenkaeser:


MirSzabo:
tested in postman with GET method and works as expected, good work!

src/server/config/swagger.json
```diff
@@ -1,11 +1,47 @@
...
+  "paths": {
+    "/crops": {
+      "get": {
+        "tags": [
+          "crops"
+        ],
+        "summary": "/api/crops",
+        "description": "Get all crops",
+        "operationId": "getAllCrops",
+        "produces": [
```
cecastosic, line 25: You already said that API can produce application/json, maybe here you can define example response, for example {
  "application/json": {
       "id": "2",
			"name": "Shiso ",
			"plant_variety": "Lamiaceae",
			"fk_user_id": "2",
			"created_at": "2020-03-16 21:03:56",
			"updated_at": "2020-03-16 21:03:56",
			"deleted_at": null
}

src/server/api/controllers/crops.controller.js
```diff
@@ -0,0 +1,13 @@
+const knex = require('../../config/db');
+
+const getAllCrops = async () => {
+  try {
+    return await knex('crops').select('crops.name', 'crops.plant_variety');
```
bytewiz, line 5: FUN FACT: you don't need to do both return AND await. It should resolve to the same thing 🚀

senner008:
@bytewiz should we merge this?

senner007:
@zenkaeser 
Can you resolve these conflicts?

ricardoaguiar:
I'm going to add the swagger-auto docs to this endpoint.

dpfernandes:
Approved

PR #71, opened 2020-03-26
Storybook : Add Crop- Details Component

mahieakhtar:
Feature add crop component

PR #106, opened 2020-04-04
Revert "Storybook : Add Crop- Details Component"

dpfernandes:
Reverts HackYourFuture-CPH/FP-class11#71

Add Crop is now called Add Batch so this needs to be reverted

@mahieakhtar will make a new PR with the changes that we need


bytewiz:
@dpfernandes @mahieakhtar If it is just the naming that changed we should just make a new Pull Request that renames it instead that will be less work IMO 👍

dpfernandes:
@bytewiz, we already have the new PR for this https://github.com/HackYourFuture-CPH/FP-class11/pull/113

Can't we just Revert and Merge PR#113?

mahieakhtar:
> @dpfernandes @mahieakhtar If it is just the naming that changed we should just make a new Pull Request that renames it instead that will be less work IMO 👍


Hi Niels,
 These are the new requirement which changed on Add Batch field:
1>Crop Name changed from input field to dropdown.
2>Deleted the Harvest Date field from the form.

dpfernandes:
This will be solved on PR #113

PR #113, opened 2020-04-04
Storybook : Initial commit for Add Batch Details

mahieakhtar:
Storybook : Initial commit for Add Batch Details


ricardoaguiar:
I have made some changes to the styling for the dropdown to better match the input fields. I have checked on chrome and firefox on windows and mac and it looks consistent.

src/client/components/form-field/form-field.component.js
```diff
@@ -1,54 +1,57 @@
...
   const handleSubmit = (e) => {
     try {
       e.preventDefault();
-      console.log(plantVariety);
-      console.log(seedPot);
+      console.log(cropName);
       console.log(startSeedDate);
-      console.log(harvestSeedDate);
+      console.log(seedPot);
       console.log(customerName);
```
MirSzabo, line 22: Hi Mahie, do not commit console.log

src/client/components/form-field/form-field.component.js
```diff
@@ -1,54 +1,57 @@
...
 import InputText from '../input-field/input-text.component';
 import InputDate from '../input-field/input-date.component';
 import Button from '../button/button.component';
 
 import './form-field.css';
 
 export default function FormField() {
-  const [plantVariety, setPlantVariety] = useState(null);
-  const [seedPot, setSeedPot] = useState(null);
+  const [cropName, setCropName] = useState(null);
```
MirSzabo, line 11: Consider using an empty string to clear the component because we have a warning in DevTools console: Warning: `value` prop on `select` should not be null.
const [cropName, setCropName] = useState('');

src/client/components/form-field/form-field.component.js
mahieakhtar: Done

src/client/components/form-field/form-field.component.js
mahieakhtar: Done

gizemcandemir:
Hi Mahie, I also made a commit to suggest using the following colors for hover effect instead of `lightgreen`. 
`input.date` and `input.input-text` classes have `#f2f2f5`. I also added active mode to their style with the same colors.
dropdown button's `.select-wrapper select` has `#f0f0f3` (because this one doesn't have inner shadow unlike others, it needed to be a little bit darker comparing others)

PR #115, opened 2020-04-05
Seeding for materials schema

mahieakhtar:
Seeding for materials schema


zkwsk:
This is included in #126

PR #117, opened 2020-04-05
firebase backend adding admin config file and middlewares authorization

### LINKED ISSUE #76: Add private routing

sowmya1408:
resolves #76


src/server/api/lib/middleware/admin/admin.js
```diff
@@ -0,0 +1,11 @@
+const admin = require('firebase-admin');
```
senner007, line 1: I suggest changing the name to firebase so that its not confused with the role

sowmya1408: @senner007 I have made the changes based on your feedback. Please review :)

PR #126, opened 2020-04-06
Feature/fix seeds

zkwsk:
This PR should fix the problems causing Travis to fail all PRs.

- Added sequencing to seed files, so they run in correct order.
- Made deleted_at column on units table nullable
- Add migration to fix mistakes in original crop stages migration, so the schema will now match the seed file for crop stages.


Ekamba:
Approve by  Axel and Mahie

PR #130, opened 2020-04-07
updated edit-batch-component deleted duplicated component

hemanth06912:
I created this PR because, edit-batch-component which is closed (issue#8) ,unfortunately that merged PR as two folders edit-batch-component and editCropImage but we need only one component that is edit-batch-component , this happened when I tried to edit the component it created a duplicate component which I didn't notice , so now I created a PR resolving the issue with as only the required component as per the requirement.


senner008:
Can you provide a description for this pr?

PR #134, opened 2020-04-07
migrations for sensors_readings and crop_stage_default_values schemas

mahieakhtar:

PR #135, opened 2020-04-07
seeding for sensors-readings and crop-stage-default-values schemas

mahieakhtar:


zkwsk:
This one can only work after #134 has been merged. That's why I suggested making a single PR for related migrations and seeds. Otherwise don't make the PR for the seed before the migration has been merged.

mahieakhtar:
Closing this PR because axel have aready created this.

PR #137, opened 2020-04-07
abstract func/end of line windows test

senner007:
bla bla bla


senner007:
Hi Sowmya
Take a look at the change.
It's easier to follow whats going on now because the sql operation is abstracted out and the operation has been named.
"Code should read like well-written prose" 
😃

This pr can just be closed, btw. Line endings doesn't seem cause any issues

senner007:
No, lets just close it.
It can just be added later with something else

PR #141, opened 2020-04-08
Firebase middlewares usage/Checks and other changes related to token

sowmya1408:
please also refer #137


src/server/api/lib/middleware/authorization.middleware.js
```diff
@@ -3,18 +3,22 @@ const { getAuthToken } = require('./getToken.backend');
...
 
+const selectUserAndRole = async (userId) => {
+  const result = await knex('users')
+    .join('user_roles', 'user_roles.fk_user_id', '=', 'users.id')
+    .join('roles', 'roles.id', '=', 'user_roles.fk_role_id')
+    .select('users.uid', 'roles.name')
+    .where('users.uid', userId);
+  return result[0];
+};
+
```
bytewiz, line 14: If you are looking for only one record from the database consider using `.first()` instead of accessing the array with an index.

Read here: http://knexjs.org/#Builder-first

src/client/firebase/getTokenWithHeaders.js
```diff
@@ -4,7 +4,7 @@ export const getTokenWithHeaders = async () => {
   const token = await Firebase.getAuth().currentUser.getIdToken();
   const headers = {
     'Content-Type': 'application/json',
-    authorization: `Sowmya ${token}`,
+    authorization: `${token}`,
```
bytewiz, line 7: Usually, the convention would be to have `Bearer ` in front of the token but doesn't seem like we are doing that so this looks fine 👍

firebase-middlewares.md
```diff
@@ -0,0 +1,58 @@
...
+const header = await getTokenWithHeaders()
+
+await fetch('http://localhost:5000/api/example-end-point', {
+method: "GET/POST",
+headers: header
+})
+.then(res => res.json())
+.then(data => console.log(data))
+.catch(err => console.log(err))
+...
```
bytewiz, line 17:
```javascript
import {getTokenWithHeaders} from '../../firebase/getTokenWithHeaders'

const header = await getTokenWithHeaders()

await fetch('http://localhost:5000/api/example-end-point', {
  method: "GET/POST",
  headers: header
})
.then(res => res.json())
.then(data => console.log(data))
.catch(err => console.log(err))
```

☝️ When you write code in a markdown file `.md` it makes it easier to read if you use the formatting feature 👍 
Like this:

image

firebase-middlewares.md
```diff
@@ -0,0 +1,58 @@
...
+In the Backend, there are middlewares for firebase(authentication.middleware & authorization.middleware) which would be used while creating Endpoints and In your routes folder/ route file it should be done as below
+
+...
+const exampleController = require('../controllers/example.controller');
+const {
+checkIfAuthorized,
+} = require('../lib/middleware/authorization.middleware');
+
+const ROLES = require('../../constants/roles');
+...
```
bytewiz, line 30: Same as above for formatting :v:

firebase-middlewares.md
```diff
@@ -0,0 +1,58 @@
...
+And then,
+
+router.get('/',checkIfAuthorized(ROLES.SUPER_ADMIN),(req, res, next) => {
+exampleController
+.getExample()
+.then((result) => res.json(result))
+.catch(next);
+});
+...
```
bytewiz, line 39: Same as above for formatting :v:

firebase-middlewares.md
zkwsk: Actually, you should be able to just insert an empty line and then indent your code with 4 spaces and it should be formatted as code.

    const example = () => "Like this";

src/server/api/lib/middleware/authorization.middleware.js
sowmya1408: changed :)

firebase-middlewares.md
```diff
@@ -0,0 +1,58 @@
...
+
+then the middleware should look like:
+
+`checkIfAuthorized(ROLES.SUPER_ADMIN,ROLE.ADMIN)`
+
+example: for Batch Details , we have SuperAdmin & Admin & User as "YES"
+
+then the middleware should look like:
+
+`checkIfAuthorized(ROLES.SUPER_ADMIN,ROLE.ADMIN,ROLE.USER)`
```
cecastosic, line 58: Shouldn't we use checkIfAuthenticated, because it doesn't matter which role is returning? It only matters that user is authenticated...

src/client/firebase/getTokenWithHeaders.js
sowmya1408: Yes, I have an error on backend with array-destructuring in eslint , so I need to change it here.Otherwise, I would used the text 'Bearer'.

firebase-middlewares.md
sowmya1408: changed :)

firebase-middlewares.md
sowmya1408: changed :)

firebase-middlewares.md
sowmya1408: Yes, you can also use checkIfAutheticated as middleware, if you also need to check the user authentication on backend and no roles needs to be specified :)

src/client/firebase/getTokenWithHeaders.js
cecastosic: @bytewiz Shouldn't we use Bearer in front of token as that is by convention?

firebase-middlewares.md
senner008: This is where a claims approach might lessen the confusion. Too late for that so nevermind
Is there a reason for having both an admin and a superadmin?

cecastosic: @senner007 I think the roles are defined based on potential future requests, and there is need to check the role only in two cases: add batch and delete batch. Please check the table for more details https://docs.google.com/spreadsheets/d/1FlBJNVvcPEJJpu-RyPz6J-m5Gto7LZ_O90LaXDkJuCs/edit#gid=0 
For everything else IMO, we need only checkIfAuthenticated. 
I would ask for more details about claims approach, but if it is too late for it... then nothing 🙂

senner008: @cecastosic I see the table yes, but Im asking if theres an actual reason to have both roles?  Cant admins not just be allowed to delete also?
Anyways... should we merge it?

cecastosic: If we don't need "Bearer" in front of the token, my answer is yes. 🙂

senner008: 👍 I'll add the changes and merge it

sowmya1408: @senner008 do you think we only need add checkifAuthorized to Add batch and delete batch and nothing else , based on the attached spreadsheet :)

PR #143, opened 2020-04-08
Endpoint to Read Role of a specific User [Issue 99]

### LINKED ISSUE #99: Endpoint to Read Role of a specific User
When I want to access a specific page, I want to be able to validate the access with my the Role that my User has so I can access that page.

Create an endpoint using nodejs, knex which can retrieve the ROLE of a specific USER. So you have to use an SELECT statement to read it from the database using knex, check database diagram for table references for which fields can be retrieved on db diagram available on the README file.

Add this endpoint to Swagger on https://github.com/HackYourFuture-CPH/FP-class11/blob/develop/src/server/config/swagger.json

cecastosic:
closes #99


senner008:
@cecastosic Isn't this branch editing the same files that we just merged in pr #141 ?

senner008:
No, they are not. Nevermind

cecastosic:
@senner007 Done 🙂 Thanks!

src/server/api/controllers/users.controller.js
```diff
@@ -0,0 +1,24 @@
...
+const getRole = async (req) => {
+  const uid = req.authId;
+  try {
+    const role = await knex('roles')
+      .join('user_roles', 'user_roles.fk_role_id', '=', 'roles.id')
+      .join('users', 'users.id', '=', 'user_roles.fk_user_id')
+      .select('roles.name')
+      .where('users.uid', uid);
+    if (role.length === 0) {
+      throw new Error(`incorrect entry with the uid of ${uid}`, 404);
```
senner007, line 14: Maybe the uid shouldn't be revealed

src/server/api/routes/users.router.js
```diff
@@ -0,0 +1,21 @@
...
+// controllers
+const usersController = require('../controllers/users.controller');
+
+const {
+  checkIfAuthenticated,
+} = require('../lib/middleware/authentication.middleware');
+
+// ENDPOINT: /api/users/role :GET to get user's role for specific uid/user
+router.get('/role', checkIfAuthenticated, (req, res, next) => {
+  usersController
```
senner007, line 15: Since we are using linter and Prettier, maybe these could be set to allow only one of the two async syntax approaches.

src/server/api/controllers/users.controller.js
cecastosic: Changed, thanks

src/server/api/routes/users.router.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-console */
+const express = require('express');
+
+const router = express.Router({ mergeParams: true });
+
+// controllers
+const usersController = require('../controllers/users.controller');
```
senner007, line 7: Also here, I think one should decide on always to either import the methods in an object or extract it as its done with the checkIfAuthenticated method

cecastosic: module.exports = {
  checkIfAuthenticated,
};
It is not used export.default with checkIfAuthenticated so it has to be imported  with {}

senner007: I'm seeing two different ways of importing here. One with es6 object destructuring and one without, right?

cecastosic:
@senner007 Can you please approve again, I fixed the package-lock.json conflict. Thanks!

sowmya1408:
It works perfect 💯

PR #144, opened 2020-04-08
Dashboard Page [Issue 92]

### LINKED ISSUE #92: Dashboard Page
When I am logged in, I want to be able to see all the components on the Dashboard page.

Please aggregate the following components:

- SideBarMenu(PR#38),
- ProgressBar(Merged)
- LineCharts(PR#35)
- Footer (Merged)

cecastosic:
closes #92


src/client/containers/dashboard-page/dashboard-page.component.js
```diff
@@ -1,123 +1,246 @@
...
-          </Card> */}
-        </div>
+    <div className="dashboard">
+      <SidebarMenu text="Dashboard" isActive={false} />
+      <div className="content">
+        <header>
+          <h1>Batch overview</h1>
+        </header>
+        <main>
+          <ProgressBar
```
senner008, line 21: This has to be decomposed. First of all it will become difficult to maintain when the function is so big. Secondly, its not descriptive. A high level component should start out by simply describing to the reader what it contains. It should take about 3-5 seconds to read.

senner008: I suggest putting all the arrays in variables. Maybe under a common object to hold all the data for the dashboard

senner008:
```
strokeGrid="#fff"
strokeAxis="#808080"
minColor="#FF0000"
maxColor="#FF0000"
optimalValue="#008000"
strokeWidthRef={2}
strokeLine="#000"
strokeWidthLine={2}
ReferanceAreaColor="#808080"
```
These properties should go in an object also. A good rule of thumb is to limit the amount of parameters that are moved around. One might easy forget to insert or add one in the function. This also makes it easier to maintain

zkwsk: So as we discussed in class, the main reason that your component is so big is that you have dummy data for the component inlined with the component itself. But since you're waiting for an endpoint to get real data that is going to be refactored anyway, so don't worry about pulling that out in separate variables.

About the changes variables Niels is referencing above, which are essentially about theming your Charts, you should have them in one central place. Normally I would put them in the theming file, but since you need the variables in js and we've not set up an easy way to read the theming file in js I would just put them together in a const containing an object in this file an reference it from there so you're not repeating yourself.

About what we discussed about having more "strict" separation of concerns between "container" and "stateless" components, I was arguing that you could add another layer do you would have a hierarchy like this:

- DashboardPageContainer ("stateful" container component)
  - DashboardPage (stateless functional component)
    - Progressbar (stateless functional component)
    - Card (stateless functional component)
    - LinechartForDashboard (stateless functional component)
    - etc.

**DashboardPageContainer**
Responsible for getting (and possibly combining data from the API and for managing state). No markup, the render function will only render the `<DashboardPage>` component. Should not be attempted mocked in storybook.

**DashboardPage**
Displays a full dashboard page. Essentially the render function will look like what you have in this component at the moment, but you will have to extract all state and put it in the container component, making this a stateless component. The responsibility of this component is to compose together other stateless components into a page layout. Should have a story in storybook.

**Any dependent child components**
This could be card components, charts or any other UI component you need to display. Should all be stateless components, although they can have some local state that is not shared with other component (i.e. disable a form field when a certain radio button is toggled or so on). They should not have side-effects, meaning that they should under no circumstances be calling the API directly. They should communicate what they need up to the container that will supply the data from the API. Should have stories in storybook.

The benefit of this approach is that you get very clear separation of concerns. Your "stateless" components will be pure functions (i.e. no side-effects) which makes them very easy to test, but more essential for us is that it makes them very easy to mock in Storybook. In your particular case you were essentially building a "story" inside the application by adding all of that mock data. If you turn your component into a stateless component you can move all of that mock data into Storybook. The benefit of doing this is that someone who needs to review the DashboardPage in it's entirety (maybe a UX designer) can go find it in Storybook without needing to have knowledge of how to get the application in a state to show that page correctly (i.e. create a user with the right priviliges, seed the right data to database).

Most projects I have worked on don't take this "pure" approach and separate the containers entirely from the markup, but personally I think it's worth-wile.

src/client/containers/dashboard-page/dashboard-page.style.css
```diff
@@ -0,0 +1,62 @@
...
+header {
+  padding: 1em 1em 0;
+}
+
+header h1 {
+  font-size: 1.3em;
+  margin: 1em 0 0;
+  padding: 0 1em;
+  text-transform: uppercase;
+}
```
zkwsk, line 25: We don't have a generic style for headers in our theme?

src/client/containers/dashboard-page/dashboard-page.style.css
```diff
@@ -0,0 +1,62 @@
...
+
+.cards .card:first-child h3 {
+  margin-bottom: 0;
+}
+
+.dashboard .modal > .content {
+  display: block;
+  height: auto;
+  margin-left: 0;
+}
```
zkwsk, line 62: This is quite well written and is mainly concerned with organizing the existing components. Good work 💪 

Note: This needs to live in the stateless `<DashboardPage>` component, not the container. So the stateless components that come with all the markup also have all of the associated styling that goes with the markup.

src/client/components/dashbord-items/dashboard-items.component.js
```diff
@@ -6,13 +6,17 @@ const DashbordItems = ({ items }) => {
   return (
     <ul className="list">
       {items.map((item) => (
-        <li key={item.id}>{item.value}</li>
+        <li key={item.id}>
+          <a href={`/chart-detail-page/${item.value.toLowerCase()}`}>
```
zkwsk, line 10: Probably want to user the `<Link>` component from React Router :)

src/client/components/logout/logout.style.css
```diff
@@ -100,9 +92,12 @@
...
   text-transform: uppercase;
-
 }
 
 .confirm-cancel:active {
   box-shadow: var(--debossed-shadow-medium);
 }
+
+.sb-show-main .popup-content {
+  position: absolute !important;
```
zkwsk, line 102: Do we really need `!important` statements already? 😱 I hope not. I'm pretty sure you can override whatever you are trying to override by adding a class or two extra. Ping me if you're in doubt.

src/client/components/dashbord-items/dashboard-items.component.js
cecastosic: I removed those changes since dashboard items (in sidebar menu) should be visible only on detail chart page and that is not part of this PR

src/client/components/logout/logout.style.css
cecastosic: Unfortunately yes, since we have to override inline style.

src/client/components/dashboard-page/dashboard-page.component.js
```diff
@@ -0,0 +1,150 @@
...
+            startDate={progressBarData.startDate}
+            currentDate={progressBarData.currentDate}
+            stages={progressBarData.stages}
+          />
+          <div className="cards">
+            <CropSummary />
+            <Card>
+              <CardTitle title="Temperature" />
+              <LineChartForDashboard
+                tempData={lineChartData.tempData}
```
senner007, line 58: But why not?
`<LineChartForDashboard lineChartData={lineChartData}/>`

senner007: It's like your saying : "Look at this code!... and btw I'm not telling your what it is"

src/client/components/line-chart-for-dashboard/line-chart-for-dashboard.component.js
```diff
@@ -11,87 +12,82 @@ import {
...
+        <XAxis
+          stroke="#666"
+          dataKey={formatXAxis}
+          domain={['dataMin', 'dataMax']}
+          tick={{ fill: '#666' }}
+        />
+        <YAxis
+          stroke="#666"
+          type="number"
+          domain={[boundary.minimum - 3, boundary.maximum + 3]}
```
MirSzabo, line 55: for charts where sensor readings numbers are not high it looks ok but if maximum is 900, 903 as end of the chart is not visible....same issue will be in the detail chart

mahieakhtar:
Please update fk_crops_id to fk_crop_id in your knex migration and controller if it's being used because fk_crops_id has been updated to fk_crop_id to make it consistent and to avoid confusion on migrations schema and DB diagram. So please replace fk_crops_id to fk_crop_id if your application is crashing and you want to save time.

zkwsk:
@cecastosic what is holding this one back? It's been going in and out of review for a long time. Is it possible to get it merged with some missing functionality and creating new PRs for adding anything you might be waiting for? It's a big chunk of the app functionality right here and the sooner we can get it into the app the better :)

cecastosic:
> @cecastosic what is holding this one back? It's been going in and out of review for a long time. Is it possible to get it merged with some missing functionality and creating new PRs for adding anything you might be waiting for? It's a big chunk of the app functionality right here and the sooner we can get it into the app the better :)

I agree 🙂 The only thing that is missing that I am aware of is status box (waiting for endpoints to be merged)...

Issue #152, opened 2020-04-10
Endpoint that Gets the Name of the user from UID


dpfernandes:
Create an endpoint using nodejs and knex that has as parameter the **UID** and returns the **Name** of the user. Check database diagram for table references for which fields can be retrieved on db diagram available on the README file.

Add this endpoint to Swagger on https://github.com/HackYourFuture-CPH/FP-class11/blob/develop/src/server/config/swagger.json

<img width="121" alt="Screenshot 2020-04-10 at 19 55 52" src="https://user-images.githubusercontent.com/6642037/79011877-50754f00-7b65-11ea-8964-4fc70bd432dd.png">

PR #155, opened 2020-04-11
Add endpoint that gets username from UID and swagger documentation

kfazinic:
solves #152

### DIFF
--- a/src/server/api/controllers/users.controller.js
+++ b/src/server/api/controllers/users.controller.js
@@ -19,6 +19,23 @@ const getRole = async (req) => {
   }
 };
 
+// get user's name for a specific uid
+const getName = async (req) => {
+  const uid = req.authId;
+  try {
+    const name = await knex('users')
+      .select('users.name')
+      .where('users.uid', uid);
+    if (name.length === 0) {
+      throw new Error(`Name not found`, 404);
+    }
+    return name;
+  } catch (error) {
+    return error.message;
+  }
+};
+
 module.exports = {
   getRole,
+  getName,
 };
--- a/src/server/config/swagger.json
+++ b/src/server/config/swagger.json
@@ -62,19 +62,23 @@
         "produces": ["application/json"],
         "parameters": [
           {
-            "name": "uid",
-            "in": "body",
-            "description": "uid from Firebase token",
+            "name": "authorization",
+            "in": "header",
+            "description": "Firebase token",
             "required": true,
-            "type": "string"
+            "type": "string",
+            "default": "Bearer ENTER_FIREBASE_TOKEN"
           }
         ],
         "responses": {
           "200": {
             "description": "Name returned."
           },
+          "401": {
+            "description": "You are not authorized to view this content."
+          },
           "404": {
-            "description": "Name not found"
+            "description": "Name not found."
           },
           "default": {
             "description": "Unexpected error."
--- a/src/server/config/swagger.json
+++ b/src/server/config/swagger.json
@@ -49,6 +49,40 @@
         ]
       }
     },
+    "/users/name": {
+      "get": {
+        "tags": ["user name"],
+        "summary": "Get a name for specific user by uid",
+        "description": "",
+        "operationId": "getName",
+        "produces": ["application/json"],
+        "parameters": [
+          {
+            "name": "uid",
+            "in": "body",
+            "description": "uid from Firebase token",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Name returned."
+          },
+          "404": {
+            "description": "Name not found"
+          },
+          "default": {
+            "description": "Unexpected error."
+          }
+        },
+        "security": [
+          {
+            "fireabse_auth": ["read"]
+          }
+        ]
+      }
+    },
     "/crop-stage-parameter-values/{cropId}": {
       "get": {
         "tags": ["crop-default-values"],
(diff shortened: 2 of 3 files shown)


src/server/api/controllers/users.controller.js
```diff
@@ -19,6 +19,23 @@ const getRole = async (req) => {
   }
 };
 
+// get user's name for a specific uid
+const getName = async (req) => {
```
senner007, line 23: How is this going o be used? By admins?

senner007: @dpfernandes 👆

cecastosic: We need it for logout popup. 
Screenshot 2020-04-11 at 18 51 02

src/server/config/swagger.json
```diff
@@ -49,6 +49,40 @@
...
+      "get": {
+        "tags": ["user name"],
+        "summary": "Get a name for specific user by uid",
+        "description": "",
+        "operationId": "getName",
+        "produces": ["application/json"],
+        "parameters": [
+          {
+            "name": "uid",
+            "in": "body",
```
cecastosic, line 62: change to header, similar to swagger in this PR https://github.com/HackYourFuture-CPH/FP-class11/pull/154

src/server/api/controllers/users.controller.js
senner007: And the name is not extractable on the client side?

kfazinic: We'll be getting the name of the currently logged-in user to display on some button in the profile :)

senner007: Ok. 
But maybe it would make more sense to only have the id in the mysql table and then the rest of the information like name and email in the firebase table...

cecastosic: Name is only in DB, in Firebase is email

senner007: Yes and is there a specific reason for having division?

cecastosic: Yes, the decision was to not use any of Firebase DB, only MYSQL.. therefore we can't add user's name to Firebase.

senner007: The Firebase authentication db is already in use. Otherwise we can't retrieve tokens from it. The username can be extracted on the client side through the property user.displayName. 
As of now I can't see why the email should be kept in mysql and the name in the firebase db.

senner007: or the other way around :)

senner007: Confusion has a tendency to build up if left unhandled.

senner007: If the name is stored in mysql when the user signs up, then we will have to make a post request and a mysql request to add the name in addition to this one to get the name. Those can both be avoided if the name is added on the client side on sign up

cecastosic: Agree Niels, but where should we add name to Firebase? This is the adding user screen: 
Screenshot 2020-04-11 at 21 38 05
And this is user.displayName
Screenshot 2020-04-11 at 21 39 46
Thanks!

senner007: Im thinking with an addition input field for the name

cecastosic: I found this https://stackoverflow.com/questions/43509021/how-to-add-username-with-email-and-password-in-firebase

senner007: I think you should simply go to the documention. It is perfectly readable

cecastosic: Sure. We don't have signup form, I am really not sure how to add user's name to Firebase. I found in the documentation how to update profile, and change displayName... but I still have to have name for it.

senner007: Not sure i follow... You shouldn't have to have a name to add a name. That would be impossible. As far as I can see it is a matter of invoking the updateProfile method

cecastosic: Niels, we don't have signup form. Which name should be written as displayName in the updateProfile method? 
user.updateProfile({
  displayName: "User's name",
})
We added users manually to Firebase, so that we can use Firebase auth. There are only name and password fields, as in the screenshot I sent you.

senner007: Well if we don't want a signup page, we would just have to have an input field where the user can set the name. You could for instance set the email as the displayName by default and then allow the user to change it later

cecastosic: Fine with me, if you think that's better. But we should communicate with @dpfernandes about it.

senner007: I'm curious, how would you have done it otherwise? Added the names manually to the mysql database?

senner007: Also, how are the users of the website going to signup, without a signup page?

cecastosic: Users table is already there, with id, uid, name, email etc

senner007: "We created a table and therefore have to use it" 
Thats not a valid argument

dpfernandes: For now we will use the hardcoded **User** values but in the future we should look into adding all the **User data on Firebase** and by doing this we will look into the **Create a new user/Signup flow.**

PR #157, opened 2020-04-11
batches endpoint added 

amjadiftikhar:


src/server/api/controllers/batches.controller.js
```diff
@@ -0,0 +1,13 @@
+const knex = require('../../config/db');
+// const Error = require('../lib/utils/http-error');
```
senner007, line 2: Remove the outcommented code

MirSzabo:
Hi Amjad endpoint works as expected tested in postman 
good work!

but for some reason I can't see all the endpoints from swagger file on page http://localhost:5000/api/documentation/
I don't have this issue on develop branch so we should check it

MirSzabo:
after update it looks ok 👍

dpfernandes:
Approved

PR #160, opened 2020-04-13
update git documentation

senner008:
This adds instructions on how to simulate a merge


senner007:
Hmm... I'm having second thoughts about this...
Maybe better to find a way to simply diff the conflicts

PR #162, opened 2020-04-13
Fix: Add 'next' to error-handling middleware

kfazinic:
Fixes `TypeError: res.sendStatus is not a function` when accessing non-existing routes.

> Define error-handling middleware functions in the same way as other middleware functions, except with four arguments instead of three, specifically with the signature (err, req, res, next)):

https://expressjs.com/en/api.html ("Error-handling middleware")


dpfernandes:
Approved

Issue #107, opened 2020-04-04
Endpoint to "DELETE" a specific BATCH (it's an UPDATE)


dpfernandes:
When I want to DELETE a BATCH, I want the system to validate my role and give access to "DELETE" the specific BATCH.

Create an endpoint using nodejs, knex which can "DELETE" a specific Batch. So you have to use an INSERT statement to UPDATE and date stamp the deleted_at field from the Batch Table using knex. Check database diagram for table references for which fields can be retrieved on db diagram available on the README file.

Add this endpoint to Swagger on https://github.com/HackYourFuture-CPH/FP-class11/blob/develop/src/server/config/swagger.json

PR #163, opened 2020-04-13
endpoints to delete specific batch

hemanth06912:
#107


src/server/api/controllers/delete-specific-batch.controller.js
```diff
@@ -0,0 +1,19 @@
+const knex = require('../../config/db');
+
+const deleteSpecificBatchById = async (batchId) => {
+  try {
+    await knex('batches_table')
```
amjadiftikhar, line 5: Is it a right name of the table? Should not be just 'batches'?

hemanth06912: yes it should be batches

ricardoaguiar:
I have fixed a couple errors on the swagger.json and the router-api files. 
I get the 401 status on swagger but could not get the 200 with a successful delete.  Is there something missing to get the successful delete status?

src/server/config/swagger.json
```diff
@@ -21,6 +21,14 @@
...
         ],
         "parameters": [
+          {
+            "name": "authorization",
+            "in": "header",
+            "description": "Firebase token",
+            "required": true,
+            "type": "string",
+            "default": "Bearer ENTER_FIREBASE_TOKEN"
+          },
```
sowmya1408, line 31: good use of firebase token object in swagger 👍

src/server/api/controllers/batch.controller.js
```diff
@@ -0,0 +1,23 @@
+const knex = require('../../config/db');
+const Roles = require('../../constants');
+
+const Error = require('../lib/utils/http-error');
+
+const deleteSpecificBatchById = async (batchId, res, next) => {
+  try {
+    const user = (await Roles.SUPER_ADMIN) === 'super_admin';
+    if (!user) return next(new Error('User have no rights to delete'));
```
sowmya1408, line 9: this code is not required, as it is taken care by the middleware checkIfAuthorized and I see that it is added to the endpoint.
Also please check there are two different files for same endpoint in controllers folder and also routes folder.

src/server/api/routes/delete-specfic-batch.router.js
```diff
@@ -0,0 +1,28 @@
...
+  checkIfAuthorized,
+} = require('../lib/middleware/authorization.middleware');
+
+const router = express.Router({ mergeParams: true });
+
+// controllers
+const deleteSpecificDataController = require('../controllers/delete-specific-batch.controller');
+
+// ENDPOINT: /api/example/ :DELETE
+router.delete('/:batchid', checkIfAuthorized(Roles.SUPER_ADMIN), (req, res) => {
```
sowmya1408, line 14: good use of the middleware  :)

src/server/api/controllers/batch.controller.js
hemanth06912: yes now it is resolved

dpfernandes:
Approved

PR #164, opened 2020-04-13
fixing routes and 404 page

cecastosic:


src/client/App.js
```diff
@@ -11,45 +12,33 @@ import PrivateRoute from './helpers/PrivateRoute';
...
         setUserState(user);
-        setAuth(true);
       } else {
         setUserState(null);
-        setAuth(false);
       }
     });
   }, []);
 
+  if (userState === undefined) return <LoaderAnimation />;
```
senner007, line 27: There are some inconsistencies here.
1. You are checking to see if the userState is undefined and later if it is null. Which is it, null or undefined?
2. LoaderAnimation shouldn't depend on the state of the user being logged in. It should depend on whether the component is loaded or not.
3. Why have the 404 set to a private route? and why is this route depending on the userState?

cecastosic: Since I am trying to use context to pass user (as userState through the whole app and all pages), I had issues with it, where when you reload the page (and the user is undefined), it will open login page and then when if it gets user from Firebase and the context, user will be set and dashboard page will be visible. LoaderAnimation is in App.js to show loading bubbles while it waits to get userState. Once you are logged out userState will be set to null. 404 page as we talked yesterday should be private route, and visible only to users that are logged in. If you are not logged in and you misspelled the route, it should show login page. Do you have suggestions how to solve it?

senner007: Well, we are dealing with two different states right?
One state to tell us if the user is logged in or not and another state to tell us if the post request to the google server has returned. I think this should be made clear.
In this case I would have the following. 
```
const [userState, setUserState] = useState(null);
const [userFetched, setUserFetched] = useState(false);
```
About the 404 being private, can you explain why that is? If the page is not found, then what does that have to do with the user being logged in or not?

senner007: I wouldn't have a state appear as both undefined, null and a value. This might result in less code, but when you return to this in a month or so, it will be a source of confusion

cecastosic: @zkwsk Can you please help with why 404 page should be private 🙂 ?
@senner007 Thanks, I made changes, please review again.

dpfernandes: @cecastosic Yes, we decided that on class. 404 is only shown for users that are already logged in. Anyone that is not logged in/authenticated should be able to understand if the page exists or not.

senner007: Well, if the user does not have a valid token, it will still be impossible to get access to the endpoints. 
In any case this should be handled entirely on the server. A middleware component should redirect to login page if unauthorized, and a 404 page otherwise.

dpfernandes:
404 page will be addressed on a different PR 
Fix route was merged via PR #166

PR #175, opened 2020-04-15
Endpoint to GET Default Values of a specific crop_stage/batch [Issue 172]

### LINKED ISSUE #172: Endpoint to GET Default Values of a specific crop_stage/batch
Make an endpoint that receives as parameters the **batch_id** and **stage** and returns the crop_stage_default_values for the crop that the batch belongs to.

If the stage parameter is **NULL**, it should return **ALL** the crop_stage_default_values
If the stage parameter is **equal to a stage** (seeding, propagation, maturity), is should return only those **specific crop_stage_default_values**

cecastosic:
closes #172


sowmya1408:
perfect 👍

PR #176, opened 2020-04-16
Endpoint to populate Status Box [Issue #142]

### LINKED ISSUE #142: Endpoint to populate Status Box
We need an endpoint that populates this Status Box

Screenshot 2020-04-12 at 13 36 44Screenshot 2020-04-12 at 13 36 36

All the operations have to assume that they get BATCH_ID as a parameter.

**Status** - need to check if all the sensors provide a value within the limits. Please follow these rules:

1. If the values from the sensors are between the max and the min it shows **Okay** .
2. If the value of more than one sensor shows values higher or lower than the max or min then the status should show **CHECK SENSORS** .
3. If it’s just one sensor that shows values higher or lower than the max or min then status should show **CHECK PH** or **CHECK WATER LEVEL** and so on.

**Days left to harvest** - Calculate the harvest start date using duration value on the crop_stages table and seeding_date value on batches table and return the number of days.

**Days left to end the batch** - Calculate the sum of all the durations of that specific batch using duration values of all the stages (crop_stages table) and seeding_date value on batches table, and return the number of days.

**Current stage** -  Using the seeding_date on the Batch Table and the duration values on Crop_stages find validate in which stage should the batch be and return the stage.

**Production Start Date** - Seeding date

**Production End Date** - From the seeding date add the number of days calculated on Days left to end the project

If any of these values are not available it should return a value NULL for that specific value.

mahieakhtar:
resolves #142 
Status label is not part of this PR.


src/server/api/routes/status-box.router.js
```diff
@@ -22,4 +22,20 @@ router.get(
   },
 );
 
+// ENDPOINT: /api/status-box/daysTilHarvest/:batchId :GET to get days till harvest date of a batch with batchId
+router.get('/days-to-harvest/:batchId', (req, res, next) => {
+  statusBoxController
```
sowmya1408, line 27: the middleware checkIfAuthenticated is missing in this endpoint, may be you can refer to the : firebase-middlewares.md document in our project to on how to use it on end point :)

src/server/api/routes/status-box.router.js
```diff
@@ -22,4 +22,20 @@ router.get(
...
+    .getDaysTilHarvest(req.params.batchId)
+    .then((result) => res.json(result))
+    .catch(next);
+});
+
+// ENDPOINT: /api/status-box/productionDate/:batchId :GET to get production date
+router.get('/production-date/:batchId', (req, res, next) => {
+  statusBoxController
```
sowmya1408, line 35: Even here the middleware is missing, just refer to above comments :)

src/server/config/swagger.json
```diff
@@ -97,42 +97,72 @@
...
-            "name": "authorization",
-            "in": "header",
-            "description": "Firebase token",
-            "required": true,
-            "type": "string",
-            "default": "Bearer ENTER_FIREBASE_TOKEN"
+            "name": "batchId",
+            "in": "path",
+            "type": "integer",
+            "required": true
```
sowmya1408: There should also be a firebase object, which give you the option to enter the token in swagger is nice to add, as it would know that you are authenticated.
`{
            "name": "authorization",
            "in": "header",
            "description": "Firebase token",
            "required": true,
            "type": "string",
            "default": "Bearer ENTER_FIREBASE_TOKEN"
          }`

src/server/api/routes/status-box.router.js
```diff
@@ -0,0 +1,25 @@
...
+  checkIfAuthenticated,
+} = require('../lib/middleware/authentication.middleware');
+
+// controllers
+const statusBoxController = require('../controllers/status-box.controller');
+
+// ENDPOINT: /api/status-box/current-stage/:batchId :GET to get current stage message of a batch with batchId
+router.get(
+  '/current-stage/:batchId',
+  checkIfAuthenticated,
```
sowmya1408, line 16: Good to see the endpoint is protected using middleware :)

src/server/config/swagger.json
```diff
@@ -202,6 +202,48 @@
...
+        "tags": ["current stage for status box"],
+        "parameters": [
+          {
+            "name": "authorization",
+            "in": "header",
+            "description": "Firebase token",
+            "required": true,
+            "type": "string",
+            "default": "Bearer ENTER_FIREBASE_TOKEN"
+          },
```
sowmya1408: 👍

src/server/api/routes/days-to-end-batch.router.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable no-console */
+const express = require('express');
+
+const router = express.Router({ mergeParams: true });
+
+// controllers
+const dayLeftController = require('../controllers/days-to-end-batch.controllers');
+
+// ENDPOINT: /api/day-left/:batchId :GET to get Days left to end the batch
+router.get('/:batchId', (req, res, next) => {
```
sowmya1408, line 10: its good to protect the endpoint with a middleware.Could you refer to firebase-middleware.md file in our project on how to use middleware on endpoint

src/server/config/swagger.json
```diff
@@ -9,6 +9,40 @@
...
+    "/days-to-end-batch/{batchId}": {
+      "get": {
+        "tags": ["status box"],
+        "parameters": [
+          {
+            "name": "batchId",
+            "in": "path",
+            "type": "integer",
+            "required": true
+          }
```
sowmya1408, line 21: An object which allows to input the firebase token should also be added to swagger json.something like this:
`{
            "name": "authorization",
            "in": "header",
            "description": "Firebase token",
            "required": true,
            "type": "string",
            "default": "Bearer ENTER_FIREBASE_TOKEN"
          },`

mahieakhtar:
Please update fk_crops_id to fk_crop_id in your knex migration and controller if it's being used because fk_crops_id has been updated to fk_crop_id to make it consistent and to avoid confusion on migrations schema and DB diagram. So please replace fk_crops_id to fk_crop_id if your application is crashing and you want to save time.

src/server/api/controllers/days-to-end-batch.controllers.js
```diff
@@ -0,0 +1,31 @@
...
+    const batch = await knex('batches')
+      .select('seeding_date', 'fk_crop_id')
+      .where('id', batchId);
+    if (batch.length === 0) {
+      throw new Error(`incorrect entry with the id of ${batchId}`, 404);
+    }
+
+    const totalDays = await knex('crop_stages')
+      .sum('duration as duration')
+      .where('fk_crops_id', batch[0].fk_crop_id);
```
mahieakhtar, line 18: Please update fk_crops_id to fk_crop_id in your knex migration and controller if it's being used because fk_crops_id has been updated to fk_crop_id to make it consistent and to avoid confusion on migrations schema and DB diagram. So please replace fk_crops_id to fk_crop_id if your application is crashing and you want to save time.

src/server/api/controllers/status-box.controller.js
```diff
@@ -0,0 +1,142 @@
...
+        'crop_stages.name',
+        'crop_stages.duration',
+        'batches.seeding_date',
+      )
+      .where('batches.id', batchId);
+    if (batchesStages.length === 0) {
+      throw new Error(`incorrect entry with the id of ${batchId}`, 404);
+    }
+
+    const cropStagedId = [1, 2, 3];
```
moonflare, line 59: Please don't use hardcoded data. 🤔

src/server/api/controllers/days-to-end-batch.controllers.js
```diff
@@ -0,0 +1,31 @@
...
+    const start = moment(batch[0].seeding_date);
+    const end = moment();
+    const daysPassed = end.diff(start, 'days');
+    const daysLeft = totalDays[0].duration - daysPassed;
+    return `${daysLeft} days`;
```
moonflare, line 23: I would return just the integer and in the frontend we add the `days` part. Also, you need to consider `1 day` case as well.

src/server/api/controllers/days-to-end-batch.controllers.js
```diff
@@ -0,0 +1,31 @@
+const knex = require('../../config/db');
+
+const Error = require('../lib/utils/http-error');
+
+const moment = require('moment');
+
+const getDayLeftToEndBatch = async (batchId) => {
+  try {
+    const batch = await knex('batches')
```
moonflare, line 9: To avoid having `batch[0]` everywhere, we can have this, which means `const batch = result[0]`, or taking the first element of the result.
```suggestion
    const [batch] = await knex('batches')
```

cecastosic:
@moonflare Thanks for your review. We refactored this PR based on your feedback, can you please review it again. 
@sowmya1408 Thanks for your review. We added the middleware, can you please review it again? 
@mahieakhtar fk_crops_id is changed to fk_crop_id - but be aware endpoints will not work, untill PR #180 is not merged to develop

mahieakhtar:
> @moonflare Thanks for your review. We refactored this PR based on your feedback, can you please review it again.
> @sowmya1408 Thanks for your review. We added the middleware, can you please review it again?
> @mahieakhtar fk_crops_id is changed to fk_crop_id - but be aware endpoints will not work, untill PR #180 is not merged to develop

@Daniel Its time to review and merge PR#180, Thank you.

moonflare:
> > @moonflare Thanks for your review. We refactored this PR based on your feedback, can you please review it again.
> > @sowmya1408 Thanks for your review. We added the middleware, can you please review it again?
> > @mahieakhtar fk_crops_id is changed to fk_crop_id - but be aware endpoints will not work, untill PR #180 is not merged to develop
> 
> @daniel Its time to review and merge PR#180, Thank you.

I can have a look at #180 later tonight.

src/server/config/swagger.json
```diff
@@ -9,6 +9,98 @@
   "host": "",
   "basePath": "/api",
   "paths": {
+    "/batch-default-values/{batchId}": {
```
moonflare, line 12: Did we add this route? or is it legacy code? I think we can remove it.

cecastosic: This is another endpoint :)

moonflare: Ah. I see it [here](https://github.com/HackYourFuture-CPH/FP-class11/blob/348853861cc9192bdacb39d89ae536be820f7f50/src/server/api/routes/api-router.js#L35). It was just added to swagger in this PR. My bad 😓

cecastosic:
> Nice work overall. But there are some performance improvements we should do before we merge this one. There is a lot of over fetching data. We fire the same queries over and over. If we call one query once, we already have that data, no need to call it again. For example the query below, we run it 3 times I think.
> 
> ```js
> knex('batches')
>       .select('seeding_date', 'fk_crop_id')
>       .where('id', batchId);
> ```
> 
> We should get all the data we need first, in one go if it makes sense, and then start calculating different metrics with that data. 🙌

Done :)

src/server/api/controllers/batch-status.controller.js
```diff
@@ -0,0 +1,152 @@
...
+    return moment(productionEndDate).format('DD-MM-YYYY');
+  } catch (error) {
+    return error.message;
+  }
+};
+
+const getBatchStatus = async (batchId) => {
+  try {
+    const batchesStages = await getBatchesStages(batchId);
+    const batch = await getBatch(batchId);
```
moonflare, line 136: Considering that we only  get the `seeding_date` from the batchs table, I think we can extract it from the `batchesStages` like so `batchesStages[0].seeding_date` as we already have that data. This way we remove one extra request to the database 👌

src/server/api/controllers/batch-status.controller.js
```diff
@@ -0,0 +1,152 @@
...
+    (batch) => batch.duration >= currentDay,
+  );
+  const currentStageName =
+    filterDurationLessThanToday.length > 0
+      ? filterDurationLessThanToday[0].name
+      : 'delivered';
+  if (currentStage.length === 0) {
+    throw new Error(`incorrect entry with the id of ${batchId}`, 404);
+  }
+  if (currentStageName === 'delivered') return { status: currentStageName };
```
moonflare, line 44: We should keep the data structure the same regardless if we have data or not. This way we only check for the values on the frontend.
```suggestion
  if (currentStageName === 'delivered') return { status: currentStageName, day: null };
```

src/server/api/controllers/batch-status.controller.js
```diff
@@ -0,0 +1,152 @@
...
+  return { status: currentStageName, day: currentDay };
+};
+
+const getDaysTilHarvest = async (batchesStages) => {
+  try {
+    const [harvestId] = await knex('crop_stages')
+      .select('id')
+      .where('name', 'harvest');
+    const durationDaysBeforeHarvest = batchesStages
+      .filter((stage) => stage.id < harvestId.id)
```
moonflare, line 54: I'm not sure I understand this filter. Can you please help me understand what is it for?

src/server/api/controllers/batch-status.controller.js
```diff
@@ -0,0 +1,152 @@
...
+        return durationSum + cropStage.duration;
+      }, 0);
+    const todayDate = moment();
+    const seedingDate = moment(batchesStages[0].seeding_date);
+    const harvestStartDate = seedingDate.add(durationDaysBeforeHarvest, 'days');
+    if (
+      moment(todayDate).format('DD-MM-YYYY') >=
+      moment(harvestStartDate).format('DD-MM-YYYY')
+    )
+      return `Harvest started`;
```
moonflare, line 65: We should avoid returning this kind of data format. If our function returns integers when data is available, we should probably return `null` when there are no days and in the frontend we know if the result is `null`, then the `Harvest started`

src/server/api/controllers/batch-status.controller.js
moonflare: Or perhaps remove `seeding_date` from `batchesStages` and only use the `seeding_date` provided by the `getBatch()` and this way we don't have the same `seeding_date` for each stage item inside `batchesStages` list. Up to you which one you pick 🙌

cecastosic: Please check now 🙂

src/server/api/controllers/batch-status.controller.js
cecastosic: Filtering all stages before harvest and summing up duration of those stages

dpfernandes:
Approved

Issue #158, opened 2020-04-12
Control authentication from env vars


zkwsk:
In order to make it easier to work with authentication we need a few more things:

* Set token time to live from env variable
* Set an "escape hatch" so you can set a default user on development environments that is always signed in
* Make sure all endpoints are protected
* Update documentation about getting token

PR #179, opened 2020-04-17
Add default user sign-in; protect routes; update token documentation

kfazinic:
Solves #158 

Regarding the 2nd task in the [issue](https://github.com/HackYourFuture-CPH/FP-class11/issues/158) (_Set an "escape hatch" so you can set a default user on development environments that is always signed in_), .env needs to be updated with 👇 

```
# To automatically sign in with a default user in development mode (npm run dev)
DEFAULT_USER=useremail@domain.com
DEFAULT_PASSWORD=password
```


dpfernandes:
Approved

PR #194, opened 2020-04-20
Export to csv

Nazaqat:


Nazaqat:
> Is this script.js tied to the project? It is sitting on the root folder of the project.
> There are also missing files from the router folder which makes npm crash when running npm run dev.
> I could not make it work.

Hi @ricardoaguiar, 
Thank you for your reviews. I think this code will be used/applied in someone else component. So the person who will apply this code needs only CODE and I ll look into router files later today evening :)

PR #202, opened 2020-04-22
New endpoint to extend the Endpoint to Get All Batches

### LINKED ISSUE #188: New endpoint to extend the Endpoint to Get All Batches
The endpoint to get all batches only returns the **ID** and the **Company_Name** and in order to populate this list we need to **create another endpoint ** with the few other new fields 

image

The new endpoint should return a list of all the batches with the following information
BATCH_ID, 
CROP_NAME, 
CUSTOMER_NAME.
STATUS,
CURRENT_STAGE,
SEEDING_POTS,
and
SEEDING_DATE

You should make sure that this endpoint is documented on SWAGER

cecastosic:
closes #188


src/server/config/swagger.json
```diff
@@ -485,74 +485,129 @@
...
+          },
+          "404": {
+            "description": "Not found"
+          },
+          "500": {
+            "description": "Server error"
           }
         }
       }
+    },
```
kfazinic, line 597: You can put this in the route file so it follows the same way @ricardoaguiar wrote the docs [here](https://github.com/HackYourFuture-CPH/FP-class11/pull/184) because I think we won't keep this file...

MirSzabo:
Hello Ceca, works as expected tested in postman 👍 great work!

MirSzabo:
http://localhost:5000/api/batches?detailed=true tested and works fine 👍

dpfernandes:
Approved

PR #150, opened 2020-04-10
sensor readings endpoint with swagger documentation  added

amjadiftikhar:

PR #204, opened 2020-04-22
export json to csv in react

Nazaqat:


src/client/containers/export-data/export-data.component.js
```diff
@@ -0,0 +1,62 @@
...
+    a.setAttribute('href', url);
+    a.setAttribute('download', 'exportedData.csv');
+    document.body.appendChild(a);
+    a.click();
+    document.body.removeChild(a);
+  };
+
+  //   Fetching data from endpoint
+  const getRequest = async function() {
+    const jsonUrl = 'https://jsonplaceholder.typicode.com/users';
```
ricardoaguiar, line 39: Can you make the jsonUrl a prop? it will be easier to share with other graphs.

Nazaqat: Hi Ricardo, Thank you for your suggestion - I ll look into that. But one other issue linked with code is, It works when I use external URL link 'https://jsonplaceholder.typicode.com/users' but I does not work on sensor reading endpoint #150  'http://localhost:3000/api/sensor-reading/' .
Component is lying in client/containers/export-data.
May its because of authentication / wrong folder placement or something else?

src/client/containers/export-data/export-data.component.js
```diff
@@ -0,0 +1,62 @@
...
+    // http://localhost:3000/api/sensor-reading/
+    const res = await fetch(jsonUrl);
+    const json = await res.json();
+
```
mahieakhtar, line 43: what if json returns empty ?

src/client/containers/export-data/export-data.component.js
mahieakhtar: There is good examples of consuming API running at your localhost https://dzone.com/articles/consuming-rest-api-with-reactjs

PR #207, opened 2020-04-23
Add Batch page - new PR

Sreelatha-Vallabhaneni:


src/client/components/add-batch-page/add-batch-page.style.css
```diff
@@ -0,0 +1,3 @@
+.add-batch {
+  margin: 0 0 0 180px;
```
zkwsk, line 2: Don't really like that we need to push in the content like this, but I see we're doing the same on other pages as well. Overall the CSS layout is not done very well with both the menu and footer being position:fixed, but I don't think we should fix this in your PR and maybe we should not touch it before wednesday.

zkwsk: Could you maybe add a TODO comment above this line, so we can find it later?

src/client/containers/add-batch-page/add-batch-page.container.js
```diff
@@ -0,0 +1,73 @@
...
+          cropId,
+          setStartSeedDate,
+          setCropId,
+          setCustomerName,
+          setSeedPot,
+          createBatch,
+          logoutModal,
+          setLogoutModal,
+        }}
+      >
```
zkwsk, line 66: Again, I don't think you should change it at this point in time, but I don't understand why you are using Context to communicate to `<AddBatchPage />` when you can just pass down props? I think I saw this pattern in another PR as well.

PR #208, opened 2020-04-23
footer-component-style-fixed

Ekamba:


src/client/components/footer/footer.style.css
```diff
@@ -6,8 +6,10 @@ footer {
 }
 
 .footer-hr {
-  border: 0;
-  height: 0;
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 100px;
```
gizemcandemir, line 13: it should be `80px;` but instead of px I suggest using `5em;`

Ekamba: I will do that 👍

Issue #189, opened 2020-04-20
List Batches Page


dpfernandes:
(right after I login), When I want to able to see a list of all the current batches so I can pick then batch that I want to monitor.

 Create a page below and use the endpoint developed on the issue to show the list https://github.com/HackYourFuture-CPH/FP-class11/issues/188

![image](https://user-images.githubusercontent.com/6642037/79752402-88725400-8314-11ea-943c-c2b5a64a1305.png)

dpfernandes:
Please fix the border of the List Batch Table. When we have the mouse over the edges should remain round.

MirSzabo:
Hello Daniel I see the table with round borders 
![table](https://user-images.githubusercontent.com/48127476/80839962-940b2800-8bfc-11ea-899f-3e2df27ad1ae.png)

PR #209, opened 2020-04-24
List batches page

MirSzabo:
solves #189
in progress: 
- sorting buttons
- filtering buttons 
- add export functionality PR #204
- add alert modal window


zkwsk:
I get redirected directly to the list-batches page, no matter what I do.

src/client/containers/login-page/login-page.component.js
```diff
@@ -37,7 +37,7 @@ function LoginPage() {
             if (err) {
               setErrorMessage(err.message);
               setErrorNotifications(err.code);
-            } else history.replace('/dashboard');
+            } else history.replace('/list-batches');
```
zkwsk, line 40: What is this about?

src/client/containers/login-page/login-page.component.js
```diff
@@ -14,7 +14,7 @@ function LoginPage() {
   const user = useContext(FirebaseContext);
 
   useEffect(() => {
-    if (user) history.replace('/dashboard');
+    if (user) history.replace('/list-batches');
```
zkwsk, line 17: What is this about?

zkwsk (changes requested):
I get redirected to the list-batches page no matter what I do. Even if I'm not logged in.

src/client/containers/login-page/login-page.component.js
zkwsk: Sorry, didn't realize that the list batches page comes up before so you can select which dashboard.

Currently you are redirected to this page even if you're not logged in though. That needs to be resolved.

Also the menu is completely blank. I think there should be options to log out and add batches (add batches, only provided you have authorization to see that).

zkwsk: Probably involves changing the sidebar component a bit to allow for easier turning on/off options.

zkwsk:
_Decided together with @dpfernandes that we will not merge this before wednesday._

Would like to set it up with a url structure so you can deeplink to the individual dashboards. This will also be necessary for the batch details component to know which batch to show details for.

In theory we could handle it with internal state in context, but this is the kind of content you typically want to be able to deeplink to.

Also, I'm not quite convinced that you should always land on this page when you sign in. If you have previously chosen a batch I think you should end on the dashboard for that batch.

PR #210, opened 2020-04-24
Fix for dashboard page

cecastosic:
Finished status box / without status label
Fixed some issues 
Fixed dashboard page without data, when stage is delivered
Fixed showing last 5 readings on charts based on date
Fixed flex issue @gizemcandemir


src/client/components/card/crop-summary-card/crop-summary-card.component.js
```diff
@@ -1,53 +1,25 @@
-import React, { useState, useEffect } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 import Card from '../card.component';
 import { CardTitle } from '../card-title.component';
-import { Status } from '../status-card/status-card.component';
+// import { Status } from '../status-card/status-card.component';
```
zkwsk, line 5: Forgotten comment?

src/client/components/card/crop-summary-card/crop-summary-card.component.js
```diff
@@ -1,53 +1,25 @@
...
+  prodEndDate,
+  stageName,
+  dayCount,
+}) => {
   return (
     <Card>
       <div className="card-header">
         <CardTitle title="STATUS" />
-        <Status stat={stat} />
+        {/* <Status stat={stat} /> */}
```
zkwsk, line 22: Comment?

src/client/containers/dashboard-page/dashboard-page.component.js
```diff
@@ -22,6 +22,7 @@ const DashboardPageContainer = () => {
   const [lineChartDataEc, setLineChartDataEc] = useState(null);
```
zkwsk, line 22: Small thing that should probably have been caight in a previous PR, but it would be a bit more intuitive if the file was called dahboard-page.container.js. If you fix it, make sure to catch any references and update them.

src/client/containers/dashboard-page/dashboard-page.component.js
```diff
@@ -54,6 +58,12 @@ const DashboardPageContainer = () => {
       },
     ).then((data) => data.json());
     setBoundaryData(defaultValues);
+
+    const statusData = await fetch('/api/batch-status/1', {
```
zkwsk, line 62: How does the batch-status endpoint work? I'm just wondering if the `1` is a hardcoded value?

src/client/components/card/crop-summary-card/crop-summary-card.component.js
cecastosic: intentionally left there to be fixed the future :)

src/client/components/card/crop-summary-card/crop-summary-card.component.js
cecastosic: intentionally left there to be fixed the future :)

sowmya1408:
Looks good 👍  
CSS fix - footer overlaps with logout button @gizemcandemir mentioned that It would be fixed. 
status box missing - but as @cecastosic mentioned based on @dpfernandes decision,we do not have it now.
These are things I found, other than that everything look goods to me :)

src/client/containers/dashboard-page/dashboard-page.component.js
cecastosic: it is hardcoded, since we are showing data only for batch with id 1, in PR #209 will be not harcoded

src/client/components/dashboard-page/dashboard-page.style.css
```diff
@@ -5,14 +5,19 @@
...
   margin-left: 130px;
   width: 100%;
 }
 
+.dashboard .content .wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  margin: 0 auto;
+  width: 1100px;
```
gizemcandemir, line 18: this should be `max-width` actually

gizemcandemir: or maybe you can just remove it, since it will be set from `index.css`

gizemcandemir:
Great job! 💯

dpfernandes:
Approved

PR #213, opened 2020-04-25
Update 404 page with new structure

gizemcandemir:
New sidebar props and logout modal components are added to 404 page. 
Page content is wrapped inside `content` and `wrapper` div's to apply our responsive design logic.
I also added the `content` and `wrapper` style to index.css as we need the same style on each page. As a result of this, I needed to change a class name in logout modal component so it's not affected by the global style.


cecastosic:
Look good! But logout function doesn't work on 404 page since it is public route, not private.

PR #220, opened 2020-04-26
made some fixes for chart-detail

sowmya1408:


Ekamba:
There is a complain here: src/client/components/detail-chart/detail-chart.stories.js#L39
Identifier 'sensors_readings' is not in camel case

PR #221, opened 2020-04-27
Fixing issues/batch details page

cecastosic:


zkwsk:
Did you fix the delete modal and then hide the delete button? 🙈 

But good job. The delete modal is not functional anyway.

PR #222, opened 2020-04-27
Fix for delete batch controller

### LINKED ISSUE #219: Bug: endpoints to delete specific batch not working
The endpoint is documented but not working even though it returns

 {
  "success": true
} 

image

Please check what was submitted on this PR https://github.com/HackYourFuture-CPH/FP-class11/pull/163 and be aware that the endpoint should NOT delete the register but it should only add the date on the field **deleted_at**
image

cecastosic:
resolves #219


dpfernandes:
Approved

PR #228, opened 2020-04-28
Get local time and update sensor_readings to UTC tz

kfazinic:
Note: need to run `db:clean` and `db:setup` locally and on the production DB after merging.


src/client/components/detail-chart/detail-chart.component.js
```diff
@@ -19,10 +19,10 @@ import {
 export default function DetailChart({ data, boundary, description, unit }) {
   const formatXAxis = () =>
     data.map((sensor) => {
+      const [date, time] = sensor.created_at.split(' ');
       // eslint-disable-next-line no-unused-vars
-      const [year, month, datetime] = sensor.created_at.slice(0, 13).split('-');
-      const [date, time] = datetime.split('T');
-      const datetimeTag = `${date}/${month} ${time}h`;
+      const [year, month, day] = date.split('-');
+      const datetimeTag = `${day}/${month} ${time.slice(0, 2)}h`;
```
zkwsk, line 25: Any reason not to use moment for formatting the date? (Excepts that it was not used before)

src/server/knexfile.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
       user: process.env.DB_USER,
       password: process.env.DB_PASSWORD,
       database: process.env.DB_NAME,
+      dateStrings: true,
```
zkwsk, line 13: What does this do? Force the DB to return strings? Why is that necessary? I would think it would be better to keep dates as timestamps for as long as possible and only render them into strings in the presentational components.

zkwsk (changes requested):

PR #234, opened 2020-05-07
Swagger update batch default values

### LINKED ISSUE #232: Review all the endpoints for documentation
When I check swagger I want to see all the endpoints that are running.

Please review if swagger is documenting all the endpoints

ricardoaguiar:
Partially resolves #232

### DIFF
--- a/src/server/api/routes/batch-default-values.router.js
+++ b/src/server/api/routes/batch-default-values.router.js
@@ -19,7 +19,10 @@ const {
  *       Get default values for specific batch by batchId
  *
  *
- *       Authentication&#58; <code>super_admin</code> <code>admin</code> <code>user</code>
+ *       Authentication&#58; <code>super_admin</code> <code>admin</code>
+ *
+ *
+ *       Authorization&#58;  <code>user</code>
  *     tags:
  *       - Default Values
  *     operationId: getDefaultValues
@@ -44,13 +47,12 @@ const {
  *          type: string
  *          enum:
  *          - current
- *          - storage
  *          - seeding
  *          - propagation
  *          - maturity
  *          - harvest
+ *          - storage
  *          - delivery
- *          default: current
  *        collectionFormat: multi
  *     security:
  *        fireabse_auth:
--- a/src/server/api/routes/batch-default-values.router.js
+++ b/src/server/api/routes/batch-default-values.router.js
@@ -16,12 +16,8 @@ const {
  *   get:
  *     summary: Get default values for specific batch
  *     description:
- *       Get default values for specific batch by batchId
- *
- *
- *       Authentication&#58; <code>super_admin</code> <code>admin</code>
- *
- *
+ *       Get default values for specific batch by batchId<br /><br />
+ *       Authentication&#58; <code>super_admin</code> <code>admin</code><br /><br />
  *       Authorization&#58;  <code>user</code>
  *     tags:
  *       - Default Values
--- a/src/server/api/routes/batch-default-values.router.js
+++ b/src/server/api/routes/batch-default-values.router.js
@@ -16,8 +16,9 @@ const {
  *   get:
  *     summary: Get default values for specific batch
  *     description:
- *       Get default values for specific batch by batchId<br /><br />
- *       Authorization&#58;  <code>user</code>
+ *       'Get default values for specific batch by batchId<br /><br />
+ *       Authentication\: <code>true</code><br /><br />
+ *       Authorization\: <code>user</code>'
  *     tags:
  *       - Default Values
  *     operationId: getDefaultValues
@@ -49,9 +50,6 @@ const {
  *          - storage
  *          - delivery
  *        collectionFormat: multi
- *     security:
- *        fireabse_auth:
- *        - read
  *     responses:
  *       200:
  *         description: Successful request
--- a/src/server/api/routes/batch-default-values.router.js
+++ b/src/server/api/routes/batch-default-values.router.js
@@ -17,7 +17,6 @@ const {
  *     summary: Get default values for specific batch
  *     description:
  *       Get default values for specific batch by batchId<br /><br />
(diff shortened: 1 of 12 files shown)


src/server/api/routes/batch-default-values.router.js
```diff
@@ -37,8 +37,21 @@ const {
...
+ *          type: string
+ *          enum:
+ *          - current
+ *          - storage
+ *          - seeding
+ *          - propagation
+ *          - maturity
+ *          - harvest
+ *          - delivery
+ *          default: current
```
cecastosic: I think by default this should be empty, not current stage

src/server/api/routes/batch-default-values.router.js
```diff
@@ -37,8 +37,21 @@ const {
...
- *        type: string
- *        required: true
+ *        description: Stage default values for a batch.
+ *        required: false
+ *        type: array
+ *        items:
+ *          type: string
+ *          enum:
+ *          - current
+ *          - storage
```
cecastosic: I would move storage after harvest, the order will be as it is in DB: seeding, propagation, maturity, harvest, storage, delivery

cecastosic (changes requested):

cecastosic:
Great job 💪 
I have one more comment as a general comment. For every endpoint there is a description for the authentication part like this:
 *Authentication: super_admin admin user
But that's related to authorization and the authorization middleware, which we use for several endpoints, but not on all. I would suggest change to:
*Authentication: user
and for endpoints that use the authorization middleware
*Authorization: super_admin admin 
or 
*Authorization: super_admin 
depends on the endpoint.

src/server/api/routes/batch-default-values.router.js
ricardoaguiar: I removed the default altogether. It has no effect on the outcome. I will leave it out.

src/server/api/routes/batch-default-values.router.js
ricardoaguiar: Fixed.

ricardoaguiar:
Do you mean to remove the `super_admin` and `admin `for the endpoints that are `user` authorized?

I have changed them to:
Authentication: `super_admin` `admin`
Authorization: `user`

Isn't a bit more clear than having just one level, i.e. either authentication or authorization?

Thanks

cecastosic:
> Authentication: `super_admin` `admin`

Authorization is related to roles (super_admin, admin...), authentication is related to is the user logged in or not. This endpoint doesn't have anything with roles ie there is no usage of authorization middleware. IMO all endpoints that don't use authorization middleware shouldn't have authorization in the description, only authentication.

src/server/api/routes/users.router.js
```diff
@@ -58,10 +57,9 @@ router.get('/role', checkIfAuthenticated, (req, res, next) => {
  * /users/name:
  *   get:
  *     summary: Get a name for specific user by uid
- *     description: Get a name for specific user by uid
- *
- *
- *       Authentication&#58; <code>super_admin</code> <code>admin</code> <code>user</code>
+ *     description: Get a name for specific user by uid <br /><br />
+ *       Authentication&#58; <code>super_admin</code> <code>admin</code><br /><br />
+ *       Authorization&#58;  <code>user</code>
```
zkwsk: Why are the colons HTML encoded? Is that really necessary? Further up you use plain colons.

zkwsk (dismissed):
Interesting discussion about Authentication/Authorization in the comments. I think at this point it is all up to you to decide how you want to document stuff like that, all  care about is that it is consistent.

But my suggestion would be something like this:

Endpoint A:
Authentication: true
Authorization: super-user, admin

Endpoint B:
Authentication: true
Authorization: super-user, admin, user

Endpoint C:
Authentication: false

To me it feels a little more in line with my understanding of authentication/authorization.

@ricardoaguiar I had a small comment about HTML encoded entities. Please check that you cannot just type regular colons, it's a tiny thing though, so I'm approving and then you can sort out how what exactly yo do about authentication annotations together :)

src/server/api/routes/users.router.js
ricardoaguiar: I tried to use normal colons during the project and found out that using regular colons inside the text description, breaks the app. I guess YAML does not like that. 

Screen Shot 2020-05-10 at 08 07 07

I just did a quick look on SO and found this solution: use quotes around the text descriptions and escape the colons with :: or \
I will implement that and remove the encodings. Good observation.

https://stackoverflow.com/questions/32991517/how-to-escape-colons-and-other-special-characters-in-a-yaml-string

https://yaml.org/spec/1.2/spec.html#id2776092

ricardoaguiar: @zkwsk 
I agree with your suggestions for Authentication and Authorization. It makes sense. 
As @cecastosic mentioned, we are not checking roles on this endpoint. But the fact that we are not checking for roles does not mean that it will work without the firebase token. I just tried and I got a 401 in return - not authorized.
That means all endpoints require authentication. The user, be either super_admin, admin or user, have to log in to generate a firebase token, which at that point can be used to access the endpoint documentation. 
I think we should implement your suggestion.

@cecastosic What do you think?

cecastosic: Once again, token is related to authentication (not authorization), and ALL endpoints are secured by authentication middleware (ie firebase token). You can choose how you want to document auth part, I am fine with @zkwsk 's solution. IMO is better to put authorization description on the endpoints where it makes sense (add batch and delete batch) and not copy/paste 
Authorization: super_admin, admin, user
on almost all endpoints. 
🙂

cecastosic (dismissed):

ricardoaguiar:
Final revision. 
We are still missing a couple of endpoints that we may have to implement at some point. 
All we need for now is to approve and merge the PR.

Thanks.

ricardoaguiar (commented):
Needs approval before merge.

